Issue #15, opened 2021-06-02
Integrate Metadata work into iSB repo


dannymandel:
Take the work that was done in https://github.com/isamplesorg/metadata/issues/17 and integrate into the iSB api.  Per discussion with @datadavev, we have settled on the following:

* add it as an option to `/thing/{identifier}`
* `format=full | core | original`

dannymandel:
This was merged/deployed last Friday.

PR #16, opened 2021-06-02
Initial API integration with transform API checked into metadata repo

dannymandel:
Details here: https://github.com/isamplesorg/isamples_inabox/issues/15


dannymandel:
@datadavev -- I'll figure out the base branch issues later.  This should merge into `develop` not `main`, but it was giving me conflicts when I did that.  I can't remember if I created the branch when mistakenly had `main` checked out.  I'll get it sorted.

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
...
+        content=content, media_type=item.resolved_media_type
     )
 
 
 @app.get(
     "/related",
     response_model=typing.List[schemas.RelationListMeta],
 )
-#async def relation_metadata(db: sqlalchemy.orm.Session = fastapi.Depends((getDb))):
+# async def relation_metadata(db: sqlalchemy.orm.Session = fastapi.Depends((getDb))):
```
dannymandel, line 138: I guess this is from running Black

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
...
 async def relation_metadata():
     """List of predicates with counts"""
-    #return crud.getPredicateCounts(db)
+    # return crud.getPredicateCounts(db)
```
dannymandel, line 141: I guess this is from running Black

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
...
     session = requests.session()
     return crud.getPredicateCountsSolr(session)
 
+
```
dannymandel, line 145: I guess this is from running Black

isb_web/main.py
```diff
@@ -176,6 +190,7 @@ async def get_related(
     return res
 '''
 
+
```
dannymandel, line 193: I guess this is from running Black

isb_web/main.py
```diff
@@ -221,6 +236,7 @@ async def get_related_solr(
         )
     return res
 
+
```
dannymandel, line 239: I guess this is from running Black

notes/solr_manage.ipynb
```diff
@@ -89,7 +89,7 @@
     "import json\n",
     "import requests\n",
     "\n",
-    "SOLR_API = \"http://localhost:8983/api/cores/isb_rel/\"\n",
+    "SOLR_API = \"http://localhost:8983/api/collections/isb_rel/\"\n",
```
dannymandel, line 92: Updating per discussion earlier today

poetry.lock
```diff
@@ -59,7 +59,7 @@ cchardet = "*"
 
 [[package]]
 name = "alembic"
-version = "1.6.2"
+version = "1.6.5"
```
dannymandel, line 62: Generally speaking how do we manage these .lock files?

pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.9"
```
dannymandel, line 13: Updated the python version because we actually do depend on 3.9 per some new type annotations.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ python-dotenv = "^0.17.1"
 networkx = "^2.5.1"
 fastapi-cache2 = "^0.1.3"
 accept-types = "^0.4.1"
+isamples_metadata = {git = "git@github.com:isamplesorg/metadata.git", rev = "develop"}
```
dannymandel, line 24: New dependency on metadata repo.

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
         raise fastapi.HTTPException(
             status_code=404, detail=f"Thing not found: {identifier}"
         )
-    if full:
+    if full or format == "full":
```
dannymandel, line 116: Wasn't sure if we wanted to delete the full parameter now that its functionality has been replaced by `format`, but I left it in place in case there were backwards compatibility issues.

isb_web/main.py
```diff
@@ -103,6 +104,7 @@ async def thing_list_types(
 async def get_thing(
     identifier: str,
     full: bool = False,
+    format: str = "original",
```
dannymandel, line 107: Can these parameters be typed to an enumeration of String values?  Do we care?

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
         raise fastapi.HTTPException(
             status_code=404, detail=f"Thing not found: {identifier}"
         )
-    if full:
+    if full or format == "full":
         return item
+    if format == "core":
+        authority_id = item.authority_id
+        if authority_id == "SESAR":
+            content = SESARTransformer(item.resolved_content).transform()
```
dannymandel, line 121: Hook in the new formatter here

isb_web/main.py
```diff
@@ -111,24 +113,36 @@ async def get_thing(
...
+        else:
+            raise fastapi.HTTPException(
+                status_code=404,
+                detail=f"Core format not available for authority_id: {authority_id}",
+            )
```
dannymandel, line 126: Is raising an exception the right thing to do in this case?

poetry.lock
datadavev: In this case it is inconsequential since I added alembic but have not yet set it up (it provided database schema migration across versions). Ideally there should be test coverage to verify that an update has no negative impact. In reality, we see if things still work and continue on.

pyproject.toml
datadavev: Yes, good catch

isb_web/main.py
datadavev: Let's leave it for now since the prototype ui uses it. Will drop this parameter once ui is switched to new format

isb_web/main.py
datadavev: They should be (makes the api clearer). Not sure exactly how to do that in fastapi, probably need to create a class of enum or something

isb_web/main.py
datadavev: It's probably a 400 error, https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

Http errors are fairly open to interpretation though, so there may be a better option.

isb_web/main.py
dannymandel: Good, I'll investigate on Friday.

poetry.lock
dannymandel: Yeah, if you view the full diff you'll see there are many many lines that changed.  Is the general workflow to run `poetry update` then let it update this file then check it in?

Maybe as part of this exercise (or shortly after the work is complete) I will write a test client that exercises the API as it is known to exist today and makes assertions against known good data.

dannymandel:
Closing this one due to creating from the wrong branch.

PR #25, opened 2021-06-18
Solr isb core records

dannymandel:
This is an initial implementation of the solr importer for the ISB core records (https://github.com//issues/23).  I haven't attempted to extensively clean up the code or look at it to see ways it could be improved.

In particular the db code in `sesar_things` was a literal copy/paste job!!! =\. Curious how you think we should generalize that bit, @datadavev, or if you think it's worth the effort.


isb_lib/core.py
```diff
@@ -40,8 +48,89 @@ def relationAsSolrDoc(
...
+                    if match is not None:
+                        doc["producedBy_samplingSite_location_elevation"] = float(match.group(1))
+                if _shouldAddMetadataValueToSolrDoc(location, "latitude") and _shouldAddMetadataValueToSolrDoc(location, "longitude"):
+                    doc["producedBy_samplingSite_location_latlon"] = f"{location['latitude']},{location['longitude']}"
+
+
+    return doc
+
 
-def solrAddRecords(rsession, records, url="http://localhost:8983/solr/isb_rel/"):
```
dannymandel, line 44: I removed this because I ran into a couple cases where I forgot to include it and it was trying to run modifications against `isb_rel` for the core records.  Could probably modularize it more than just forcing it onto callers, but I haven't thought about how yet.

isb_lib/core.py
```diff
@@ -40,8 +48,89 @@ def relationAsSolrDoc(
     doc["tstamp"] = datetimeToSolrStr(ts)
     return doc
 
+def _shouldAddMetadataValueToSolrDoc(metadata: typing.Dict, key: typing.AnyStr) -> bool:
+    shouldAdd = False
+    if key in metadata:
+        value = metadata[key]
+        if type(value) is list:
```
dannymandel, line 55: Attempted to not index empty fields.  If you can think of additional type-specific checks we should be doing here, please fire away!

isb_lib/core.py
```diff
@@ -6,13 +6,21 @@
...
+import re
 
 SOLR_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
-
+ELEVATION_PATTERN = re.compile(r"\s*(-?\d+\.?\d*)\s*m?", re.IGNORECASE)
 
 def getLogger():
-    return logging.getLogger("isb_lib.core")
+    logger = logging.getLogger("isb_lib.core")
+    logger.level = logging.DEBUG
```
dannymandel, line 22: I need to fix this up.  Will do before I merge.

scripts/sesar_things.py
```diff
@@ -305,7 +306,7 @@ def _commit(session, relations):
 
     L = getLogger()
     rsession = requests.session()
-    batch_size = 5000
+    batch_size = 1
```
dannymandel, line 309: Need to double check this before I merge.

isb_lib/core.py
datadavev: That's fine. Better to force a choice when there's multiple options

isb_lib/core.py
datadavev: Could perhaps add a range check for lat lon values, but otherwise this seems fine. I presume strings are stripped of leading and trailing whitespace earlier?

datadavev (approved):

isb_lib/core.py
dannymandel: Actually, we aren't doing that.  Sounds like we should.  Easy enough to add, I can do that.

dannymandel: Implemented both suggestions, please take a look at latest version.

datadavev (approved):

isb_lib/core.py
```diff
@@ -65,7 +63,15 @@ def _shouldAddMetadataValueToSolrDoc(metadata: typing.Dict, key: typing.AnyStr)
     shouldAdd = False
     if key in metadata:
         value = metadata[key]
-        if type(value) is list:
+        if key =='latitude':
+            shouldAdd = -90.0 <= value <= 90.0
+            if not shouldAdd:
+                getLogger().error("Invalid latitude %f", value)
+        elif key == 'longitude':
```
datadavev, line 70: Just a note that longitude is an odd one to check for range since it's cyclical. Technically any value can work since it's multiples of 360. -180 ... 180 is a general convention though, and this works.

datadavev (approved):

isb_lib/core.py
datadavev: Looks good.

PR #34, opened 2021-07-08
Opencontext things

dannymandel:
Initial implementation of https://github.com//issues/32

Doesn't handle reloading existing records in the database, but does handle the initial import

### DIFF
--- a/isb_lib/opencontext_adapter.py
+++ b/isb_lib/opencontext_adapter.py
@@ -109,8 +109,11 @@ class OpenContextRecordIterator(isb_lib.core.IdentifierIterator):
                 # raise NotImplementedError
                 yield record
                 num_records += 1
-            self.url = data["next-json"]
+            self.url = data.get("next-json")
             if len(data.get("oc-api:has-results", {})) < _page_size:
+                more_work = False
+            elif self.url is None:
+                # If we hit the last page, there won't be a 'next-json' key
                 more_work = False
             elif 0 < self._max_entries <= num_records:
                 more_work = False
--- a/isb_lib/opencontext_adapter.py
+++ b/isb_lib/opencontext_adapter.py
@@ -1,0 +1,85 @@
+import datetime
+import isb_lib.core
+
+import logging
+import requests
+
+HTTP_TIMEOUT = 10.0  # seconds
+OPENCONTEXT_API = "https://opencontext.org/subjects-search/.json?add-attribute-uris=1&attributes=obo-foodon-00001303%2Coc-zoo-has-anat-id%2Ccidoc-crm-p2-has-type%2Ccidoc-crm-p45-consists-of%2Ccidoc-crm-p49i-is-former-or-current-keeper-of%2Ccidoc-crm-p55-has-current-location%2Cdc-terms-temporal%2Cdc-terms-creator%2Cdc-terms-contributor&prop=oc-gen-cat-sample-col%7C%7Coc-gen-cat-bio-subj-ecofact%7C%7Coc-gen-cat-object&response=metadata%2Curi-meta&sort=updated--desc"
+
+
+def get_logger():
+    return logging.getLogger("isb_lib.opencontext_adapter")
+
+
+class OpenContextIdentifierIterator(isb_lib.core.IdentifierIterator):
+    def __init__(
+        self,
+        offset: int = 0,
+        max_entries: int = -1,
+        date_start: datetime.datetime = None,
+        date_end: datetime.datetime = None,
+    ):
+        super().__init__(
+            offset=offset,
+            max_entries=max_entries,
+            date_start=date_start,
+            date_end=date_end,
+        )
+
+    def records_in_page(self):
+        L = get_logger()
+        L.debug("records_in_page")
+        url = OPENCONTEXT_API
+        headers = {
+            "Accept": "application/json",
+        }
+        _page_size = 200
+        params = {
+            "rows": _page_size,
+        }
+        more_work = True
+        while more_work:
+            response = requests.get(
+                url, params=params, headers=headers, timeout=HTTP_TIMEOUT
+            )
+            if response.status_code != 200:
+                L.error(
+                    "Unable to load records; status: %s; reason: %s",
+                    response.status_code,
+                    response.reason,
+                )
+                break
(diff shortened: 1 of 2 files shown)


dannymandel:
@datadavev, haven't put much polish on this one yet, but it works and it's running now on Mars.

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+        t_created: datetime.datetime,
+        status: int,
+        resolved_url: str,
+        t_resolved: datetime.datetime,
+        resolve_elapsed: float,
+        media_type: str = None,
+    ) -> igsn_lib.models.thing.Thing:
+        L = get_logger()
+        L.debug("OpenContextItem.asThing")
+        # Note: SESAR incorrectly returns "application/json;charset=UTF-8" for json-ld content
```
dannymandel, line 38: Obviously need to remove this comment.

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+            tresolved=t_resolved,
+            resolve_elapsed=resolve_elapsed,
+        )
+        if not isinstance(self.item, dict):
+            L.error("Item is not an object")
+            return _thing
+        _thing.item_type = "sample"
+        _thing.related = None
+        _thing.resolved_media_type = media_type
+        # _thing.resolve_elapsed = resolve_elapsed
```
dannymandel, line 57: Right now the way this is hooked up, I can't really wire this in.  If it's important it's certainly possible, but it looked like it was convenient for SESAR because we had a secondary http request to get details for a particular record.  Since we get them all at once in opencontext, we don't have access to the raw http response at the call site.

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+            date_start=date_start,
+            date_end=date_end,
+            page_size=page_size,
+        )
+        self.url = OPENCONTEXT_API
+
+    def records_in_page(self):
+        L = get_logger()
+        L.debug("records_in_page")
+        headers = {"Accept": "application/json", "user-agent": "isamplesbot-3000/0.0.1"}
```
dannymandel, line 83: Strangely, setting the `user-agent` header on the request is what calmed the rate limiter down.  It was certainly unexpected, but it unblocked us and it isn't harmful, so…🤷‍♂️

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+
+        After completion, self._cpage contains the next page of entries or None if there
+        are no more pages available, and self._page_offset is set to the first entry (usually 0)
+
+        """
+        is_none = self._cpage is None
+        if is_none or self._page_offset >= len(self._cpage):
+            if not is_none:
+                # Work around the rate limiter
+                time.sleep(0.3)
```
dannymandel, line 141: Not ideal, but it should ensure that we are executing roughly 3 requests/second.

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+    Args:
+        thing_dict: Dictionary representing the thing
+        t_resolved: When the item was resolved from the source
+
+    Returns:
+        Instance of Thing
+    """
+    L = get_logger()
+    id = thing_dict["uri"]
+    t_created = dateparser.parse(thing_dict.get("published"))
```
dannymandel, line 168: Another difference here -- OpenContext records don't have a creation date the way that GEOME does, so we use the published date as a best placeholder.

scripts/opencontext_things.py
```diff
@@ -0,0 +1,155 @@
...
+    L = get_logger()
+    records = isb_lib.opencontext_adapter.OpenContextRecordIterator(
+        max_entries=max_count, date_start=start_from, page_size=200
+    )
+
+    num_ids = 0
+    for record in records:
+        L.info("got next id from open context %s", record)
+        num_ids += 1
+        id = record["uri"]
```
dannymandel, line 52: The `uri` field seemed most ID-like of everything in the OpenContext records.

scripts/opencontext_things.py
```diff
@@ -0,0 +1,155 @@
...
+    #             record = next(records)
+    #             print("got next id from open context %s", record)
+
+
+def load_open_context_entries(session, max_count, start_from=None):
+    loop = asyncio.get_event_loop()
+    future = asyncio.ensure_future(
+        _load_open_context_entries(session, max_count, start_from=start_from)
+    )
+    loop.run_until_complete(future)
```
dannymandel, line 91: I'm not sure there's any point to running this in asyncio as it's currently written, but it was here initially due to it starting life as fork of the GEOME importer.

scripts/opencontext_things.py
```diff
@@ -0,0 +1,155 @@
...
+    "--verbosity",
+    default="DEBUG",
+    help="Specify logging level",
+    show_default=True,
+)
+@click.option(
+    "-H", "--heart_rate", is_flag=True, help="Show heartrate diagnositcs on 9999"
+)
+@click_config_file.configuration_option(
+    config_file_name="/Users/mandeld/iSamples/isamples_inabox/isb.cfg"
```
dannymandel, line 116: I will clean this bit up!

isb_lib/opencontext_adapter.py
```diff
@@ -0,0 +1,175 @@
...
+                    response.reason,
+                )
+                break
+            # L.debug("recordsInProject data: %s", response.text[:256])
+            data = response.json()
+            for record in data.get("oc-api:has-results", {}):
+                L.info("records_in_page Record id: %s", record.get("uri", None))
+                # print(json.dumps(record, indent=2))
+                # raise NotImplementedError
+                yield record
```
dannymandel, line 109: I don't quite understand why we use `yield` here.

isb_lib/opencontext_adapter.py
dannymandel: After setting the `user-agent` and running things again without the sleep, things are fine.  We take a bit of time to do the db inserts, so I don't think we need this after all.  It seems like the `user-agent` was the key.

dannymandel:
Merging to develop as this appears to be working on mars.cyverse.org

PR #36, opened 2021-07-23
Changes to import OpenContext records into our solr index

dannymandel:
After examining the OpenContext records, from the solr point of view they were effectively identical to the SESAR import, so I refactored things a bit to make things easer to import more collections and eliminate some of the copy pasta we had previously.

### DIFF
--- a/isb_lib/core.py
+++ b/isb_lib/core.py
@@ -10,9 +10,11 @@ import typing
 import isamples_metadata.Transformer
 
 import igsn_lib.time
+import igsn_lib.models
 from isamples_metadata.Transformer import Transformer
 import dateparser
 import re
+import requests
 
 RECOGNIZED_DATE_FORMATS = [
     "%Y",  # e.g. 1985
@@ -21,14 +23,15 @@ RECOGNIZED_DATE_FORMATS = [
     "%Y-%m-%d %H:M:%S",  # e.g 2019-12-08 15:54:00
 ]
 DATEPARSER_SETTINGS = {
-    'DATE_ORDER': 'YMD',
-    'PREFER_DAY_OF_MONTH': 'first',
-    'TIMEZONE': 'UTC',
-    'RETURN_AS_TIMEZONE_AWARE': True
+    "DATE_ORDER": "YMD",
+    "PREFER_DAY_OF_MONTH": "first",
+    "TIMEZONE": "UTC",
+    "RETURN_AS_TIMEZONE_AWARE": True,
 }
 
 SOLR_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 ELEVATION_PATTERN = re.compile(r"\s*(-?\d+\.?\d*)\s*m?", re.IGNORECASE)
+
 
 def getLogger():
     return logging.getLogger("isb_lib.core")
@@ -59,15 +62,16 @@ def relationAsSolrDoc(
     doc["tstamp"] = datetimeToSolrStr(ts)
     return doc
 
+
 def _shouldAddMetadataValueToSolrDoc(metadata: typing.Dict, key: typing.AnyStr) -> bool:
     shouldAdd = False
     value = metadata.get(key)
     if value is not None:
-        if key =='latitude':
+        if key == "latitude":
             shouldAdd = -90.0 <= value <= 90.0
             if not shouldAdd:
                 getLogger().error("Invalid latitude %f", value)
-        elif key == 'longitude':
+        elif key == "longitude":
             shouldAdd = -180.0 <= value <= 180
             if not shouldAdd:
                 getLogger().error("Invalid longitude %f", value)
@@ -78,6 +82,7 @@ def _shouldAddMetadataValueToSolrDoc(metadata: typing.Dict, key: typing.AnyStr) 
         else:
             shouldAdd = True
     return shouldAdd
+
 
 def coreRecordAsSolrDoc(coreMetadata: typing.Dict) -> typing.Dict:
     """
@@ -124,6 +129,7 @@ def coreRecordAsSolrDoc(coreMetadata: typing.Dict) -> typing.Dict:
 
     return doc
 
+
 def handle_curation_fields(coreMetadata: typing.Dict, doc: typing.Dict):
     curation = coreMetadata["curation"]
     if _shouldAddMetadataValueToSolrDoc(curation, "label"):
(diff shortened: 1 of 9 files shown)


isb_lib/core.py
```diff
@@ -21,15 +23,16 @@
...
 ]
 DATEPARSER_SETTINGS = {
-    'DATE_ORDER': 'YMD',
-    'PREFER_DAY_OF_MONTH': 'first',
-    'TIMEZONE': 'UTC',
-    'RETURN_AS_TIMEZONE_AWARE': True
+    "DATE_ORDER": "YMD",
+    "PREFER_DAY_OF_MONTH": "first",
+    "TIMEZONE": "UTC",
+    "RETURN_AS_TIMEZONE_AWARE": True,
```
dannymandel, line 29: It looked like Black hadn't run on the file previously, so there are a few formatting changes.  This is one of them.

isb_lib/core.py
```diff
@@ -59,16 +62,18 @@ def relationAsSolrDoc(
...
+
 def _shouldAddMetadataValueToSolrDoc(metadata: typing.Dict, key: typing.AnyStr) -> bool:
     shouldAdd = False
     value = metadata.get(key)
     if value is not None:
-        if key =='latitude':
-            shouldAdd = -90.0 <= value <= 90.0
+        if key == "latitude":
+            # Explicitly disallow bools as they'll pass the logical test otherwise and solr will choke downstream
+            shouldAdd = type(value) is not bool and -90.0 <= value <= 90.0
```
dannymandel, line 72: One of the interesting things we found in the OpenContext data was that there was a longstanding bug where they had things like 
```
"latitude": False,
"longitude": False
```
 in lieu of omitting them from the data.  Interestingly, the previous check that made assertions about whether the lat/lon values were reasonable passed because `False` coerced down to 0 so it was between -90 and 90.  Had to tighten up the logic to omit Booleans from the test.  @ekansa is working on a fix in the OpenContext codebase but this workaround fixes it in the meantime.

isb_lib/core.py
```diff
@@ -170,10 +178,16 @@ def handle_produced_by_fields(coreMetadata: typing.Dict, doc:typing.Dict):
                 location_str = location["elevation"]
                 match = ELEVATION_PATTERN.match(location_str)
                 if match is not None:
-                    doc["producedBy_samplingSite_location_elevationInMeters"] = float(match.group(1))
-            if _shouldAddMetadataValueToSolrDoc(location, "latitude") and _shouldAddMetadataValueToSolrDoc(location,
-                                                                                                           "longitude"):
-                doc["producedBy_samplingSite_location_latlon"] = f"{location['latitude']},{location['longitude']}"
```
dannymandel, line 176: Black formatter

isb_lib/core.py
```diff
@@ -215,15 +232,16 @@ def solrAddRecords(rsession, records, url):
...
     else:
         L.debug("Successfully posted data %s to url %s", str(data), str(_url))
 
+
 def solrCommit(rsession, url):
     L = getLogger()
-    headers = {"Content-Type":"application/json"}
-    params = {"commit":"true"}
+    headers = {"Content-Type": "application/json"}
+    params = {"commit": "true"}
```
dannymandel, line 244: Black formatter

isb_lib/core.py
```diff
@@ -339,3 +357,114 @@ def listFormats(self, profile=None):
 
     def listProfiles(self, format=None):
         return []
+
```
dannymandel, line 360: These two classes were refactorings of previous SESAR code.

isb_lib/opencontext_adapter.py
```diff
@@ -148,6 +149,21 @@ def _getPage(self):
...
+    if not thing.authority_id == OpenContextItem.AUTHORITY_ID:
+        raise ValueError("Thing is not an OpenContext item")
+
+def reparse_as_core_record(thing: igsn_lib.models.thing.Thing) -> typing.Dict:
+    _validate_resolved_content(thing)
+    try:
+        transformer = isamples_metadata.OpenContextTransformer.OpenContextTransformer(thing.resolved_content)
+        return isb_lib.core.coreRecordAsSolrDoc(transformer.transform())
+    except Exception as e:
+        get_logger().fatal("Failed trying to run transformer on %s", str(thing.resolved_content))
```
dannymandel, line 164: The lat/lon False values was one major problem we encountered, and there was one more.  It turns out a number of the OpenContext records didn't have persistent identifiers, which caused the transformer to blow up.  Per discussion with @ekansa, this is not an expected valid scenario so our assumption that all records would have a persistent is correct.  In the meantime, we log the exception and skip the record during import.  Once it's fixed, we will need to fetch those records and rerun the solr importer.  It ends up being ~2200 records out of > 800k, so most are ok.

isb_lib/sesar_adapter.py
```diff
@@ -20,7 +20,11 @@
 import isamples_metadata.SESARTransformer
 
 HTTP_TIMEOUT = 10.0  # seconds
+##############
+# Per Lulin, we shouldn't be using this as they are wanting to deprecate it
```
dannymandel, line 24: Unrelated comment, but I was making changes and wanted to make sure the code was documented that we shouldn't be using this anymore.

scripts/geome_things.py
```diff
@@ -347,28 +347,6 @@ def reloadRecords(ctx, status_code):
...
-            params["offset"] = offset
-            params["limit"] = page_size
-            qry = session.execute(sql, params)
-            for rec in qry:
-                n += 1
-                yield rec
-            if n == 0:
-                break
-            offset += page_size
-
```
dannymandel, line 371: Deleting duplicate code, one of the happiest parts of my day.

scripts/opencontext_things.py
```diff
@@ -147,11 +147,26 @@ def main(ctx, db_url, verbosity, heart_rate):
...
+    db_url = ctx.obj["db_url"]
+    solr_importer = isb_lib.core.CoreSolrImporter(
+        db_url=db_url,
+        authority_id=isb_lib.opencontext_adapter.OpenContextItem.AUTHORITY_ID,
+        db_batch_size=1000,
+        solr_batch_size=1000,
+        solr_url="http://localhost:8983/api/collections/isb_core_records/",
+    )
+    allkeys = solr_importer.run_solr_import(isb_lib.opencontext_adapter.reparse_as_core_record)
+    L.info(f"Total keys= {len(allkeys)}")
```
dannymandel, line 169: After the refactoring, this is effectively all the code it takes to run a solr import for a new provider.  Not much!

tests/test_core.py
```diff
@@ -53,11 +59,65 @@ def test_coreRecordAsSolrDoc():
...
+      },
+      "placeName": [
+        "International Space Station",
+        "Zvezda Service Module"
+      ]
+    }
+  },
+  "relatedResource": [],
+  "sampleidentifier": "ark:/28722/k2ks6xw6t"
+}
```
dannymandel, line 117: This is an example test for the broken lat/lon case.

dannymandel:
@datadavev going to merge this one so I may proceed on to Smithsonian.  Please leave any feedback on here and I will address in followup PRs.

PR #59, opened 2021-08-31
Get tests to run as part of github

dannymandel:
Set up tests like the other repo


dannymandel:
@datadavev FYI -- we now have tests running for iSB PRs.

PR #61, opened 2021-09-01
Set up integration test to run against live server

dannymandel:
Test this out…


dannymandel:
@datadavev FYI -- setting up a scheduler to run an integration test against mars so we know if the API is broken

PR #64, opened 2021-09-15
Initial Docker experiment

### LINKED ISSUE #69: Run all the things in Docker
We want to run and deploy everything in a Docker container.  Make the necessary changes to get this working.

dannymandel:
Things start up in Docker with this configuration.  There is no data in the database or solr, so we'll need to figure out the right approach there.

The containing directory looks like this:

```
(python-docker) mandeld@SBS-7448 python-docker % ls -la
total 8
drwxr-xr-x   4 mandeld  staff   128 Sep 15 16:14 .
drwxr-xr-x+ 53 mandeld  staff  1696 Sep 15 16:06 ..
-rw-r--r--@  1 mandeld  staff   485 Sep 15 16:05 docker-compose.yml
lrwxr-xr-x   1 mandeld  staff    39 Sep 15 14:18 isamples_inabox -> /Users/mandeld/iSamples/isamples_inabox
```

The contents of `docker-compose.yml` are as follows:

```
version: "3.8"
services:
    db:
        image: postgres
        environment:
            - POSTGRES_PASSWORD=isamplesinabox
            - POSTGRES_USER=isb_writer
            - POSTGRES_DB=isb_1
        volumes:
            - pgdata:/var/lib/postgresql/data
        
    solr:
        image: solr:latest
        volumes:
            - solrdata:/var/solr
        ports:
            - "8983:8983"
    isamples_inabox:
        build: ./isamples_inabox/
        ports:
            - "8000:8000"
        depends_on:
            - solr
            - db
    solr_schema_init:
        build: ./isamples_inabox/
        depends_on:
            - isamples_inabox
            - db
            - solr
        command: sleep 10 && python solr_schema_init/create_isb_core_schema.py    
    
volumes:
    pgdata:
    solrdata:
```

I'm able to open http://localhost:8000/docs in Safari and hit the API, etc after running `docker-compose up --build`.


Dockerfile
```diff
@@ -0,0 +1,10 @@
+# syntax=docker/dockerfile:1
+FROM python:latest
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip3 install -r requirements.txt
+
+COPY . .
+CMD [ "uvicorn", "isb_web.main:app", "--host", "0.0.0.0", "--port", "8000"]
```
dannymandel, line 10: Starts our app up on port 8000.

isb_web/config.py
```diff
@@ -1,9 +1,9 @@
 from pydantic import BaseSettings
 
 class Settings(BaseSettings):
-    database_url: str = "UNSET"
+    database_url: str = "postgresql+psycopg2://isb_writer:isamplesinabox@db/isb_1"
     web_root: str = "/"
-    solr_url: str = "UNSET"
+    solr_url: str = "http://solr:8983/solr/isb_core_records"
```
dannymandel, line 6: I need to figure out why the `.env` file wasn't working here.  I knew I could make it work by hardcoding these for now, and the hostnames match up with the values provided in the containing docker compose file.  Mainly just wanted to make sure that the services could see one another, and they were definitely able to do that.

isb_web/main.py
```diff
@@ -52,9 +52,9 @@
     fastapi.staticfiles.StaticFiles(directory=os.path.join(THIS_PATH, "static")),
     name="static",
 )
-templates = fastapi.templating.Jinja2Templates(
-    directory=os.path.join(THIS_PATH, "templates")
-)
+# templates = fastapi.templating.Jinja2Templates(
+#     directory=os.path.join(THIS_PATH, "templates")
+# )
```
dannymandel, line 57: Had to comment out all the Jinja stuff, due to 

```
isamples_inabox_1  |   File "/usr/local/lib/python3.9/site-packages/starlette/templating.py", line 52, in __init__
isamples_inabox_1  |     assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
isamples_inabox_1  | AssertionError: jinja2 must be installed to use Jinja2Templates
```

I didn't investigate or try to fix yet, as I didn't need it to verify that things came up correctly.

isb_lib/core.py
```diff
@@ -342,12 +344,15 @@ def solr_max_source_updated_time(
     }
     _url = f"{url}select"
     res = rsession.get(_url, headers=headers, params=params)
-    dict = res.json()
-    docs = dict["response"]["docs"]
-    if docs is not None and len(docs) > 0:
-        return dateparser.parse(docs[0]["sourceUpdatedTime"])
-    else:
-        return None
+    try:
```
dannymandel, line 347: This is a bug that has always been here if you were to try and run against an empty index.  This was added after we had data in the index, so we never hit it before.

pyproject.toml
```diff
@@ -24,6 +24,8 @@ accept-types = "^0.4.1"
 isamples_metadata = {git = "https://github.com/isamplesorg/metadata.git", branch = "develop"}
 shapely = "^1.7.1"
 geojson = "^2.5.0"
+Jinja2 = "^3.0.1"
+heartrate = "^0.2.1"
```
dannymandel, line 28: Had to add these two lines since poetry has a bug where it misses transitive dependencies.  Explicitly adding them to the list makes the `requirements.txt` export correctly.

scripts/geome_things.py
```diff
@@ -147,6 +147,9 @@ def loadGEOMEEntries(session, max_count, start_from=None):
 @click.option(
     "-d", "--db_url", default=None, help="SQLAlchemy database URL for storage"
 )
+@click.option(
+    "-s", "--solr_url", default=None, help="Solr index URL"
+)
```
dannymandel, line 152: You'll see these same changes in all the bootstrapping/boilerplate `things` scripts.  If we didn't have anything else to do, I think we could unify all the scripts in a much cleaner way.

solr_schema_init/create_isb_core_schema.py
```diff
@@ -1,9 +1,13 @@
...
 import typing
-
 import requests
+import time
+
+# Sleep for 20 seconds to allow for the Solr Docker container to start up.
+time.sleep(20)
 
-SOLR_API = "http://localhost:8983/api/collections/isb_core_records/"
+CREATE_CORE_API = "http://solr:8983/admin/cores?action=CREATE&name=isb_core_records"
```
dannymandel, line 9: I never got this part working.  I'll comment it out and add some notes.  I might try messing with it for a bit more, too, as it would be nicer if we didn't have to go through the admin UI to do this.

dannymandel:
Also take a look at the Docker repo, which will use these changes to start things up in a container: https://github.com/isamplesorg/isamples_docker

isb_web/config.py
dannymandel: Figured this out.

isb_web/main.py
dannymandel: Figured this out.  Poetry bug.

PR #70, opened 2021-09-29
Geome partial update

### LINKED ISSUE #66: Adopt GEOME API that will let us only import recently modified records
Per discussion with @jdeck88, GEOME has an API that should let us get only recently modified records.

> https://api.geome-db.org/projects/1/expeditions gives a list of mod dates per expedition for each project, and the query for > records per expedition is: https://api.geome-db.org/v1/records/Sample/jsonlimit=10&page=0&q=_projects_:1&_expedition_:auxroc_CR_aj

We should investigate how we can switch over to this API in lieu of the one we're using now.

dannymandel:
Update the Geome import to use their API that has modify dates so we can properly implement partial updates.

### DIFF
--- a/isb_lib/geome_adapter.py
+++ b/isb_lib/geome_adapter.py
@@ -55,79 +55,6 @@ def geomeEventRecordTimestamp(record):
 
 
 class GEOMEIdentifierIterator(isb_lib.core.IdentifierIterator):
-    comparison_date = datetime.datetime(year=2021, month=5, day=15)
-    known_existing_projects = [
-    "255",
-    "293",
-    "241",
-    "78",
-    "118",
-    "319",
-    "286",
-    "384",
-    "383",
-    "190",
-    "191",
-    "344",
-    "305",
-    "244",
-    "71",
-    "316",
-    "70",
-    "373",
-    "354",
-    "13",
-    "83",
-    "67",
-    "365",
-    "64",
-    "243",
-    "328",
-    "309",
-    "284",
-    "382",
-    "323",
-    "154",
-    "1",
-    "212," "74",
-    "295",
-    "357",
-    "222",
-    "292",
-    "75",
-    "242",
-    "317",
-    "240",
-    "201",
-    "348",
-    "44",
-    "282",
-    "339",
-    "181",
-    "73",
-    "248",
-    "72",
-    "85",
-    "337",
-    "90",
-    "79",
-    "82",
-    "69",
-    "80",
-    "81",
-    "68",
-    "40",
-    "254",
-    "61",
-    "223",
-    "41",
-    "76",
-    "221",
-    "290",
-    "77",
-    "89",
-    "168",
-    ]
     def __init__(
         self,
         offset: int = 0,
@@ -195,10 +122,9 @@ class GEOMEIdentifierIterator(isb_lib.core.IdentifierIterator):
             if type(expedition_dict) is dict:
                 expedition_modified_date = expedition_dict["modified"]
                 expedition_modified_datetime = dateparser.parse(expedition_modified_date)
-                have_data_for_project = str(project_id) in self.known_existing_projects
                 # For each expedition, check to see if the mod date is greater than the last time we fetched.
                 # If it is, fetch the bcids of the samples in the expedition.
-                if expedition_modified_datetime.timestamp() > self._date_start.timestamp() or not have_data_for_project:
+                if expedition_modified_datetime.timestamp() > self._date_start.timestamp():
                     more_work = True
                     params = {
                         "limit": _page_size,
--- a/isb_lib/geome_adapter.py
+++ b/isb_lib/geome_adapter.py
@@ -172,7 +172,6 @@ class GEOMEIdentifierIterator(isb_lib.core.IdentifierIterator):
         L.debug("recordsInProject project %s", project_id)
         params = {"includePublic": "true", "admin": "false"}
         url = f"{GEOME_API}projects/{project_id}/expeditions"
-        #url = f"{GEOME_API}records/{record_type}/json"
         headers = {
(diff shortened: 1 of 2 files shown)


isb_lib/geome_adapter.py
```diff
@@ -54,6 +55,79 @@ def geomeEventRecordTimestamp(record):
 
 
 class GEOMEIdentifierIterator(isb_lib.core.IdentifierIterator):
+    comparison_date = datetime.datetime(year=2021, month=5, day=15)
+    known_existing_projects = [
```
dannymandel: Wanted to get this checked in so we have a record of it for historical purposes, however, this is not part of the plan moving forward.  This list was generated by this SQL query:

```
select distinct(resolved_content->'record'->'projectId') from Thing where authority_id='GEOME';
```

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
     def recordsInProject(self, project_id, record_type):
         L = getLogger()
         L.debug("recordsInProject project %s", project_id)
-        url = f"{GEOME_API}records/{record_type}/json"
+        params = {"includePublic": "true", "admin": "false"}
+        url = f"{GEOME_API}projects/{project_id}/expeditions"
```
dannymandel, line 101: This is the main change -- after we hit a project, iterate over all the expeditions in the project and only fetch the records for those expeditions that have mod dates after the mod date we had previously.

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
...
-                # raise NotImplementedError
-                yield record
-            if len(data.get("content", {}).get(record_type, [])) < _page_size:
-                more_work = False
-            params["page"] = params["page"] + 1
+            return
+        # L.debug("recordsInProject data: %s", response.text[:256])
+        expeditions_json = response.json()
+        if len(expeditions_json) == 0:
+            L.error("Project doesn't have any expeditions project:%s", project_id)
```
dannymandel, line 119: This is mostly for our debugging purposes -- a number of projects don't have any expeditions.  From our perspective there isn't a lot to do here, but it's worth recording this.

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
...
+        for expedition_dict in expeditions_json:
+            # Skip developer messages
```
dannymandel, line 121: In some of the cases, we were getting strings back instead of dicts -- just double check that it's a dictionary before treating it as such.

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
...
+            if type(expedition_dict) is dict:
+                expedition_modified_date = expedition_dict["modified"]
+                expedition_modified_datetime = dateparser.parse(expedition_modified_date)
+                have_data_for_project = str(project_id) in self.known_existing_projects
```
dannymandel: This is the check we described this morning, @datadavev -- it lets us know if the project is one of the ones that we've identified as behaving badly.  I'll use this list to generate the report we talked about so we may write a proper bug report for @jdeck88.

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
...
+                # For each expedition, check to see if the mod date is greater than the last time we fetched.
+                # If it is, fetch the bcids of the samples in the expedition.
+                if expedition_modified_datetime.timestamp() > self._date_start.timestamp(): # or not have_data_for_project:
```
dannymandel: Only fetch details for expeditions whose mod date is greater than our last recorded mod date.

isb_lib/geome_adapter.py
```diff
@@ -96,39 +170,68 @@ def listProjects(self):
...
+                                expedition_dict["expeditionCode"],
+                                response.status_code,
+                                response.reason,
+                            )
+                            break
+                        # L.debug("recordsInProject data: %s", response.text[:256])
+                        expeditions_json = response.json()
+
+                        for record in expeditions_json.get("content", {}).get(record_type, []):
+                            L.debug("recordsInProject Record id: %s", record.get("bcid", None))
```
dannymandel, line 152: This piece is somewhat interesting in that we're now recording the mod date of the records as the mod date of the expedition.  Otherwise, we could end up refetching things.

isb_lib/geome_adapter.py
```diff
@@ -150,13 +253,11 @@ def _loadEntries(self):
...
             if not p["loaded"]:
-                for record in self.recordsInProject(p["project_id"], self._record_type):
+                for (record, modified_date) in self.recordsInProject(p["project_id"], self._record_type):
                     # record identifier
                     rid = record.get("bcid", None)
-                    # record timestamp
-                    t_collected = geomeEventRecordTimestamp(record)
                     if not rid is None:
-                        p["identifiers"].append((rid, t_collected))
+                        p["identifiers"].append((rid, modified_date))
```
dannymandel, line 184: This is the change to store the modified date of the expedition rather than the collection time (what we had previously).

isb_lib/geome_adapter.py
```diff
@@ -297,7 +398,7 @@ def reparseThing(thing):
     return thing
 
 
-def loadThing(identifier, t_created):
+def loadThing(identifier: str, t_created: datetime.datetime, existing_thing: Thing) -> Thing:
```
dannymandel, line 325: Previously, we weren't properly dealing with existing records and we'd just skip them.  This change does the overwriting as expected.

dannymandel:
@jdeck88 this is more of an FYI for you but feel free to do a full review if you're interested.  Thanks!

PR #73, opened 2021-10-06
Alternate identifiers

### LINKED ISSUE #68: iSamples API missing child Thing records, while they're available in Solr
Per discussion with @datadavev:

> One thing I’ve noticed is that we have a disparity between solr records and “thing” records for GEOME and OpenContext sources.

This is because we are generating core records for sub-records, which is what was intended, but causes a situation where the identifier in solr can’t be used to get the record from the database using the API.

For example, the GEOME record ark:/21547/CYR2envbio08_98 is in solr and can be access via the API at thing/ark:/21547/CYR2envbio08_98. However the child of that record, ark:/21547/CYT2MBIO16729.1 can be found by solr, but not through the API.

This is a fairly fundamental issue with the content representation differences across the collections, and something that the “relations” collection in solr can be used for.

Option are basically: 1) store each of the record components as separate records in the things collection (so that GEOME would have an a sample and a tissue record) or 2) provide a lookup to the parent record by following relations (e.g. o:ark\:\/21547\/CYT2MBIO16729\.1 returns the source record id).

Both approaches have issues, but I think #1 provides the simplest way to get consistent representation across the collections. This would mean that for GEOME records, we would have Event, Sample, and Tissue records in things. The Event would be used to populate the location and time collected information for the Sample and Tissue core records that get added to Solr. Conceptually, both the “Sample” and “Tissue” records are different types of physical samples, and they share a common collection event. This is similar to how the SESAR records are presented (with their parent / child records).

dannymandel:
Wire in support for alternate identifiers so we may look up GEOME child records and OpenContext records by their publication identifiers.  Still need to test it live as part of the import process.

### DIFF
--- a/isb_web/sqlmodel_database.py
+++ b/isb_web/sqlmodel_database.py
@@ -7,7 +7,7 @@ from sqlalchemy import Index
 from sqlalchemy.exc import ProgrammingError
 from sqlmodel import SQLModel, create_engine, Session, select
 
-import isb_lib.core
+import isb_lib
 from isb_lib.models.thing import Thing, ThingIdentifier
 from isb_web.schemas import ThingPage
 
@@ -211,7 +211,7 @@ def _insert_geome_identifiers(session: Session, thing: Thing):
 def _insert_open_context_identifiers(session: Session, thing: Thing):
     citation_uri = thing.resolved_content["citation uri"]
     if citation_uri is not None and type(citation_uri) is str:
-        open_context_uri = isb_lib.core.normalized_id(citation_uri)
+        open_context_uri = isb_lib.normalized_id(citation_uri)
         open_context_identifier = ThingIdentifier(
             guid=open_context_uri, thing_id=thing.primary_key
         )
--- a/isb_web/sqlmodel_database.py
+++ b/isb_web/sqlmodel_database.py
@@ -5,7 +5,7 @@ from typing import Optional, List
 from sqlalchemy import Index
 from sqlalchemy.exc import ProgrammingError
 from sqlmodel import SQLModel, create_engine, Session, select
-from isb_lib.models.thing import Thing, Identifier
+from isb_lib.models.thing import Thing, ThingIdentifier
 from isb_web.schemas import ThingPage
 
 
@@ -57,7 +57,7 @@ class SQLModelDAO:
         self._create_index(resolved_status_authority_id_idx)
 
         guid_thing_id_idx = Index(
-            "guid_thing_id_idx", Identifier.guid, Identifier.thing_id
+            "guid_thing_id_idx", ThingIdentifier.guid, ThingIdentifier.thing_id
         )
         self._create_index(guid_thing_id_idx)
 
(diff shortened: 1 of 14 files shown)


isb_web/sqlmodel_database.py
```diff
@@ -101,7 +107,12 @@ def read_things_summary(
 
 def get_thing_with_id(session: Session, identifier: str) -> Optional[Thing]:
     statement = select(Thing).filter(Thing.id == identifier)
-    return session.exec(statement).first()
+    result = session.exec(statement).first()
+    if result is None:
+        # Fall back to querying the Identifiers table
+        join_statement = select(Thing).join(ThingIdentifier).where(ThingIdentifier.guid == identifier)
+        result = session.exec(join_statement).first()
```
dannymandel, line 118: These are the 4 lines or so that actually implement the feature.  The rest is all to support this 🥇

isb_web/sqlmodel_database.py
```diff
@@ -171,3 +182,68 @@ def get_sample_types(session: Session):
...
+        _insert_geome_identifiers(session, thing)
+    elif thing.authority_id == "OPENCONTEXT":
+        _insert_open_context_identifiers(session, thing)
+    _insert_standard_identifier(session, thing)
+
+
+def delete_identifiers(session: Session, thing: Thing) -> bool:
+    if thing.primary_key is None:
+        # No pk, hasn't been saved
+        return False
```
dannymandel, line 237: This is a no-op if we don't have any saved.

isb_web/sqlmodel_database.py
```diff
@@ -171,3 +182,68 @@ def get_sample_types(session: Session):
...
+    else:
+        identifier_select = select(ThingIdentifier).where(ThingIdentifier.thing_id == thing.primary_key)
+        thing_identifiers = session.exec(identifier_select).all()
+        for identifier in thing_identifiers:
+            session.delete(identifier)
```
dannymandel, line 244: I realize I could have done this more efficiently by writing in raw SQLAlchemy, but this was the way the SQLModel docs recommended we do it.  I'm open to writing it the other way, but I wanted to follow the docs as much as possible.

scripts/update_identifiers_for_existing_things.py
```diff
@@ -0,0 +1,102 @@
+from time import sleep
```
dannymandel, line 1: This was the script that backfilled the data for our existing Things db.  I ran it on mars today and all was well.

dannymandel: (Except it made the integration test against Mars fail during the time it was running -- interesting note for the performance problems we've seen there…)

tests/test_sqlmodel_database.py
```diff
@@ -106,4 +107,85 @@ def test_thing_iterator(session: Session):
     for thing in iterator.yieldRecordsByPage():
         assert type(thing) is Thing
         count_iterated_things += 1
-    assert num_things == count_iterated_things
\ No newline at end of file
+    assert num_things == count_iterated_things
+
```
dannymandel, line 145: These are all tests for the new database functionality.  It's been really nice being able to write unit tests for all the stupid typos I make and have them caught during the unit testing phase rather than when we run it on mars.

dannymandel:
One more small piece of work -- need to write the piece that will choose the proper transformer for GEOME.  I suspect this will be 4-5 lines of code, but it is unimplemented as of yet.

dannymandel:
See also https://github.com/isamplesorg/metadata/pull/77

PR #76, opened 2021-10-19
Script to clean up OpenContext identifiers that erroneously included the n2t.net prefix.

### LINKED ISSUE #74: Update Solr record for OpenContext to use id consistent with postgres records

dannymandel:
Fix for

https://github.com/isamplesorg/isamples_inabox/issues/74

Don't really *need* to commit this as it's already run and fixed everything but I like having things there for reference and documentation purposes


dannymandel:
@datadavev This more of an FYI than anything…

PR #79, opened 2021-10-20
Create an option to rebuild the full solr index

dannymandel:
Needed to fully rebuild OpenContext with new metadata mapping, didn't have a way to do that.


dannymandel:
@datadavev FYI

PR #81, opened 2021-10-22
Initial sitemap generation

### LINKED ISSUE #77: Generate sitemap for content
sitemap.xml documents provide an index to content available on a site and can include properties such as timestamps and content type information for each location advertised.

Implement a mechanism to generate a set of sitemaps (there are limits to the site of a sitemap) based on the current content available. The target content is the iSamples core records.

dannymandel:
Initial sitemap generation implementation.  Currently doesn't handle the case where there's more than one core record per row in the Things db.  However, general code works and outputs valid xml that conforms to the schema.


tests/test_utils.py
```diff
@@ -0,0 +1,25 @@
+import datetime
+
+from sqlmodel import Session
+
+from isb_lib.models.thing import Thing
```
dannymandel, line 5: @datadavev not sure what you think about this approach for shared code for unit tests.  Please let me know if you have a better/cleaner way to approach this.

tests/test_thing_sitemap.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+from sqlmodel import Session, SQLModel, create_engine, select
+from sqlmodel.pool import StaticPool
+
+from isb_lib.sitemaps.thing_sitemap import SitemapIndexIterator
+from test_utils import _add_some_things
```
dannymandel: Would also appreciate some eyes on the relative import of the module in the same directory here.  When I initially refactored the code out to the different module, PyCharm wanted to import it as `tests.test_utils` even though the test files were in the `tests` directory, and it broke when I tried to run them.  I manually changed it to omit the leading `tests.` and things worked correctly.  It also ran correctly on GitHub, so I think it's just a PyCharm bug but if you could confirm that'd be helpful.

datadavev: I've had this same issue, never really tracked it down all the way. I think pycharm may make some assumptions on the relative paths based on how folders are marked in the UI.

tests/test_utils.py
datadavev: I think that's the right way. At least, it aligns with what I've done in the past.

Just a note - when there's a chance of namespace confusion I tend to use the full path when referencing imports. So like:

```python
import isb_lib.models.thing

something = isb_lib.models.thing.Thing()
```
It's more verbose but helps when tracking down stuff in larger files. Not an issue in this case.

PR #92, opened 2021-11-02
Initial implementation of sitemap integration tester

### LINKED ISSUE #89: Make sitemap tester that resolves the sitemap
We should set up a GitHub action that hits the sitemap URL and crawls it to verify all the URLs are resolvable.

dannymandel:
Integration test that parses out the sitemap and verifies we can fetch things from the API.

Resolves #89.

### DIFF
--- a/integration_tests/test_sitemaps.py
+++ b/integration_tests/test_sitemaps.py
@@ -1,0 +1,85 @@
+import pytest
+import requests
+import os
+import lxml.etree
+import random
+import dateparser
+import gzip
+from lxml.etree import _Element
+
+
+@pytest.fixture
+def rsession():
+    return requests.session()
+
+
+@pytest.fixture
+def sitemap_index_url():
+    sitemap_index_url = os.getenv("INPUT_SITEMAP_INDEX_URL")
+    if sitemap_index_url is None:
+        sitemap_index_url = "https://mars.cyverse.org/sitemaps/sitemap-index.xml"
+    return sitemap_index_url
+
+
+def _assert_date_tag_text(date_tag: _Element):
+    last_mod_date = dateparser.parse(date_tag.text)
+    assert last_mod_date is not None
+
+
+def test_sitemap(rsession: requests.sessions, sitemap_index_url: str):
+    """
+    General approach is as follows:
+    (1) Hit the sitemap-index.xml, and choose one of the random sitemaps inside
+    (2) Download the sitemap and unzip it
+    (3) Choose 100 of the random urls in the sitemap and resolve the URL
+    (4) Also, for all dates, verify that they are parseable by dateparser
+    """
+    res = rsession.get(sitemap_index_url)
+    xmlp = lxml.etree.XMLParser(
+        recover=True,
+        remove_comments=True,
+        resolve_entities=False,
+    )
+    root = lxml.etree.fromstring(res.content, parser=xmlp)
+    sitemap_list = root.getchildren()
+    """These sitemap children look like this:
+          <sitemap>
+            <loc>http://mars.cyverse.org/sitemaps/sitemap-5.xml.gz</loc>
+            <lastmod>2006-08-10T12:00:00Z</lastmod>
+          </sitemap>    
+    """
+    random_child_index = random.randrange(0, len(sitemap_list) - 1)
+    random_child = sitemap_list[random_child_index]
+    random_child_children = random_child.getchildren()
+    for child_element in random_child_children:
+        if "lastmod" in child_element.tag:
+            _assert_date_tag_text(child_element)
+        elif "loc" in child_element.tag:
+            sitemap_file_loc = child_element.text
+            res = rsession.get(sitemap_file_loc)
+            uncompressed_text = gzip.decompress(res.content)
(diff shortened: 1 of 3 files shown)


integration_tests/test_sitemaps.py
```diff
@@ -0,0 +1,85 @@
...
+    random_child_index = random.randrange(0, len(sitemap_list) - 1)
+    random_child = sitemap_list[random_child_index]
+    random_child_children = random_child.getchildren()
+    for child_element in random_child_children:
+        if "lastmod" in child_element.tag:
+            _assert_date_tag_text(child_element)
+        elif "loc" in child_element.tag:
+            sitemap_file_loc = child_element.text
+            res = rsession.get(sitemap_file_loc)
+            uncompressed_text = gzip.decompress(res.content)
```
dannymandel: @datadavev This exercise ended up being pretty straightforward once I figured out `lxml`, but this line surprised me.  The documentation indicated `requests` was supposed to handle gzip decompression automatically, but I found that to not be the case.  It's unclear to me if that's a bug in this code, a bug in our server, or if this is behaving as expected.  Would love your eyes on that line, please.

dannymandel:
Should add that this is running as part of our integration test suite, so we should have continuous coverage on our deployed sitemaps.

integration_tests/test_sitemaps.py
datadavev: Requests needs a hint from the server to determine what to do with it. Content that is compressed on the fly by the server includes a `Transfer-Encoding: gzip` header [^1], and requests will handle that automatically. 

In this case, the content is being served as a gzip file, so requests is simply transferring the requested resource (i.e. a gzipped file). Requests may handle decompression if we configure nginx to include the `Transfer-Encoding` header for those files. 

As an aside, the server should be configured to at least set the `Content-Encoding: gzip` header to make it clear to clients that the resource is gzipped [^2].

[^1]: https://www.rfc-editor.org/rfc/rfc2616#section-14.41
[^2]: https://www.rfc-editor.org/rfc/rfc2616#section-14.11

datadavev: I reconfigured nginx a little:
```
        gzip            on;
        gzip_types      text/plain application/xml;
```
This provides on-the-fly compression to clients that support it. For example:

```
curl -H "Accept-Encoding: gzip, deflate, br" -v "https://mars.cyverse.org/sitemaps/sitemap-index.xml"
...
> GET /sitemaps/sitemap-index.xml HTTP/1.1
> Host: mars.cyverse.org
> User-Agent: curl/7.64.1
> Accept: */*
> Accept-Encoding: gzip, deflate, br
>
< HTTP/1.1 200 OK
< Server: nginx/1.14.0 (Ubuntu)
< Date: Wed, 03 Nov 2021 15:14:59 GMT
< Content-Type: application/xml
< Transfer-Encoding: chunked
< Connection: keep-alive
< last-modified: Wed, 03 Nov 2021 05:54:35 GMT
< Content-Encoding: gzip
<
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
```

Requesting without the `Accept-Encoding` request header returns the regular XML:
```
curl -v "https://mars.cyverse.org/sitemaps/sitemap-index.xml"
...
> GET /sitemaps/sitemap-index.xml HTTP/1.1
> Host: mars.cyverse.org
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: nginx/1.14.0 (Ubuntu)
< Date: Wed, 03 Nov 2021 15:16:53 GMT
< Content-Type: application/xml
< Content-Length: 16781
< Connection: keep-alive
< last-modified: Wed, 03 Nov 2021 05:54:35 GMT

<?xml version="1.0" encoding="utf-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 https://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd">
  <sitemap>
    <loc>http://mars.cyverse.org/sitemaps/sitemap-0.xml.gz</loc>
    <lastmod>2006-03-22T12:00:00Z</lastmod>
  </sitemap>
  <sitemap>
    <loc>http://mars.cyverse.org/sitemaps/sitemap-1.xml.gz</loc>
    <lastmod>2006-03-24T12:00:00Z</lastmod>
  </sitemap>
...
```

dannymandel: So that is for the unzipped sitemap index file, but the place where I had to manually unzip was the gzipped individual sitemap files.  If I'm understanding you right, I misunderstood the Requests documentation -- it doesn't automatically gunzip actual .gz files, only those that have been compressed by effectively `mod_gzip` style compression, which is what it looks like we now have on nginx.  Is that right?

dannymandel: Per discussion in this morning's standup, we will change the sitemap generation code to emit URLs without the `.gz`, and rely on nginx to serve up the gzipped version using its raw configuration.

PR #93, opened 2021-11-03
Merge metadata into isb

### LINKED ISSUE #85: Move metadata repo code into iSB
We should move all the python code in the metadata repo into the iSB repo.

dannymandel:
Import the previous metadata repo into isb.  I suspect this might break some stuff but it is working locally…


dannymandel:
The individual files don't require review, they are just straight copies of the files in the metadata repo.  The only thing that actually changed were some values in the unit testing code as those files moved to a destination directory that I thought made sense for this project.  Note that we'll need to make some changes on mars when we deploy this change as we'll need to delete the project dependency from the virtualenv.  I made the same change with poetry locally and it seemed to work fine -- at least PyCharm was hitting breakpoints in the integrated source files rather than the ones in the virtualenv.

Issue #91, opened 2021-11-02
Bug in OpenContext fetcher where we wipe the resolved_url


dannymandel:
There is a bug in the OpenContext updater that knocked out the `resolved_url` field causing one of the automated integration tests to fail.

```
select * from thing where id='http://opencontext.org/subjects/f5a17d69-7354-43e9-bf78-d415b6ee0320';
                                  id                                  |            tstamp             |        tcreated        | item_type | authority_id | related | log | resolved_url | resolved_status |           tresolved           | resolve_elapsed |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         resolved_content                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | resolved_media_type |   _id   
----------------------------------------------------------------------+-------------------------------+------------------------+-----------+--------------+---------+-----+--------------+-----------------+-------------------------------+-----------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------+---------
[...]

PR #96, opened 2021-11-03
Incorrectly setting resolved_url to NULL for Opencontext

dannymandel:
Bug fix for https://github.com/isamplesorg/isamples_inabox/issues/91

### DIFF
--- a/isb_lib/opencontext_adapter.py
+++ b/isb_lib/opencontext_adapter.py
@@ -76,6 +76,7 @@ class OpenContextRecordIterator(isb_lib.core.IdentifierIterator):
             date_end=date_end,
             page_size=page_size,
         )
+        self._past_date_start = False
         self.url = OPENCONTEXT_API
 
     def records_in_page(self):
@@ -88,7 +89,7 @@ class OpenContextRecordIterator(isb_lib.core.IdentifierIterator):
         }
         more_work = True
         num_records = 0
-        while more_work and self.url is not None:
+        while more_work and self.url is not None and not self._past_date_start:
             L.info("trying to hit %s", self.url)
             response = requests.get(
                 self.url, params=params, headers=headers, timeout=HTTP_TIMEOUT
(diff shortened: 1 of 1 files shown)


isb_lib/opencontext_adapter.py
```diff
@@ -76,6 +76,7 @@ def __init__(
             date_end=date_end,
             page_size=page_size,
         )
+        self._past_date_start = False
```
dannymandel, line 79: New field to catch the situation we were in where we've reached the end of iteration due to being past the start date.

isb_lib/opencontext_adapter.py
```diff
@@ -113,21 +115,23 @@ def records_in_page(self):
                         self._date_start
                     )
                     data = {}
+                    self._past_date_start = True
                     break
                 # print(json.dumps(record, indent=2))
                 # raise NotImplementedError
                 yield record
                 num_records += 1
-            self.url = data.get("next-json")
```
dannymandel, line 121: This was the bug here -- previously we reached the end of the records because we were past the start date, and we cleared the current data, but then asked for the next url from the empty data, which set it to null.

isb_lib/opencontext_adapter.py
```diff
@@ -113,21 +115,23 @@ def records_in_page(self):
...
+            if (
+                len(data.get("oc-api:has-results", {})) < _page_size
+                or self.url is None
+                or 0 < self._max_entries <= num_records
+                or num_records == self._page_size
+                or self._past_date_start
+            ):
                 more_work = False
+            if more_work:
+                self.url = next_url
```
dannymandel, line 134: Now we only advance the url if there are more records to iterate.

isb_lib/opencontext_adapter.py
```diff
@@ -113,21 +115,23 @@ def records_in_page(self):
...
-                more_work = False
-            elif num_records == self._page_size:
+
+            if (
+                len(data.get("oc-api:has-results", {})) < _page_size
+                or self.url is None
+                or 0 < self._max_entries <= num_records
+                or num_records == self._page_size
+                or self._past_date_start
+            ):
```
dannymandel, line 131: This was just cleanup as the if-statement kept growing and it had the same 1-line statement in 4 places.

dannymandel:
@datadavev the only thing I haven't fully resolved is why this just started happening.  AFAICT, it should have been broken every time the updater ran and updated an existing record if we hit the end of the list.

dannymandel:
In my local db, it was definitely present:

```
isb_2=> select count(*) from thing where authority_id='OPENCONTEXT' and resolved_url is NULL;
 count 
-------
   239
(1 row)
```

PR #103, opened 2021-11-10
Initial batch of solr tests

dannymandel:
A few initial tests for querying by provider and by some of the search terms we found in providers logs

### DIFF
--- a/integration_tests/test_solr_search_results.py
+++ b/integration_tests/test_solr_search_results.py
@@ -1,0 +1,103 @@
+import pytest
+import requests
+import typing
+import os
+
+
+@pytest.fixture
+def rsession():
+    return requests.session()
+
+
+@pytest.fixture
+def solr_url():
+    solr_url = os.getenv("INPUT_SOLR_URL")
+    if solr_url is None:
+        solr_url = "https://mars.cyverse.org/thing/select"
+    return solr_url
+
+
+def headers():
+    return {
+        "accept": "application/json",
+        "User-Agent": "iSamples SOLR Integration Bot 2000",
+    }
+
+
+authority_ids = [
+    "SESAR",
+    "OPENCONTEXT",
+    "GEOME",
+    "SMITHSONIAN",
+]
+
+
+def _send_solr_query(rsession: requests.sessions, solr_url: str, query: str) -> typing.List[typing.Dict]:
+    params = {
+        "start": 0,
+        "limit": 10,
+        "fl": "*",
+        "q": f"searchText:{query}",
+    }
+    res = rsession.get(solr_url, headers=headers(), params=params)
+    response_dict = res.json()
+    docs = response_dict.get("response").get("docs")
+    return docs
+
+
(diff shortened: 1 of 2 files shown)


integration_tests/test_solr_search_results.py
```diff
@@ -0,0 +1,103 @@
...
+    rsession: requests.sessions, solr_url: str, authority_id: str
+):
+    docs = _send_solr_query(rsession, solr_url, authority_id)
+    for doc in docs:
+        appears_in_search_text = _value_in_search_text(authority_id, doc)
+        assert appears_in_search_text
+
+
+opencontext_projects = [
+    "Çatalhöyük Zooarchaeology",
```
dannymandel, line 70: @datadavev when we discussed this one this morning, I though it was failing due to some character encoding issues.  Thankfully, it turned out that wasn't the case -- there are actually two opencontext projects with roughly this title, and solr is preferring the other one.  Given that both terms (Çatalhöyük, Zooarchaeology) appear in `searchText`, I loosened the check and treated this as a pass.

integration_tests/test_solr_search_results.py
```diff
@@ -0,0 +1,103 @@
...
+            appears_in_search_text = _value_in_search_text(word, doc)
+            assert appears_in_search_text
+            assert doc["source"] == "OPENCONTEXT"
+
+
+geome_search_terms = [
+    "SYMBIOCODE",
+    "INDO_PIRE",
+    "PEER_2016",
+    "gastropoda",
```
dannymandel, line 96: I went through the geome logs that @jdeck88 sent over, and these all appeared in the search terms, and thankfully all are currently returning geome records.

scripts/solr_reindex_template.py
```diff
@@ -31,9 +31,11 @@ def main(ctx):
         if mutated_record is not None:
             current_mutated_batch.append(mutated_record)
         if len(current_mutated_batch) == batch_size:
+            logging.info(f"Going to save {len(current_mutated_batch)} records")
```
dannymandel, line 34: These two lines are unrelated, but were in my branch from previous reindexing tasks.

integration_tests/test_solr_search_results.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def _send_solr_query(
+    rsession: requests.sessions, solr_url: str, query: str
+) -> typing.List[typing.Dict]:
+    params = {
+        "start": 0,
+        "limit": 10,
+        "fl": "*",
+        "q": f"searchText:({query})",
```
dannymandel, line 42: @datadavev, after getting some unexpected results, I changed the solr query to be parenthesized and that caused us to get better results in a couple cases (New Zealand, United States).  Let me know if you think that is a reasonable default.

dannymandel: Though even with the parentheses, "Excavations at Polis" still didn't work as expected.

PR #116, opened 2022-01-12
Flake8

### LINKED ISSUE #115: Enable flake8 linter
We should adopt the flake8 linter and run it on every pull request.

dannymandel:
Try to enable flake8 initially and on every PR moving forward


dannymandel:
flake8 is running on all PRs now!  And…it passes!

isb_lib/opencontext_adapter.py
```diff
@@ -161,16 +160,18 @@ def _getPage(self):
...
 
+
 def reparse_as_core_record(thing: isb_lib.models.thing.Thing) -> typing.List[typing.Dict]:
     _validate_resolved_content(thing)
     try:
         transformer = isamples_metadata.OpenContextTransformer.OpenContextTransformer(thing.resolved_content)
         return [isb_lib.core.coreRecordAsSolrDoc(transformer)]
     except Exception as e:
-        get_logger().error("Failed trying to run transformer on %s", str(thing.resolved_content))
+        get_logger().error(f"Failed trying to run transformer on {str(thing.resolved_content)}, exception: {e}")
```
datadavev: I think % formatting should be used in log messages as opposed to fstrings because the fstrings are computed even if the log level is insufficient to emit the message, whereas % formatting is not.

Also, I don't thing the str() is needed as the logger should do that - if the message is to be emitted

datadavev (changes requested):
minor suggestion on log message formatting

isb_lib/opencontext_adapter.py
dannymandel: Good point on the log messages, I had forgotten that.

dannymandel: And I wrote a test to make sure that the formatting works as expected:

```
dict = {
    "foo": "bar"
}
logging.error("This is a dict error: %s", dict)

try:
    raise AttributeError("This is an attribute error")
except Exception as e:
    logging.error("The exception was %s", e)
```

yields:

```
ERROR:root:This is a dict error: {'foo': 'bar'}
ERROR:root:The exception was This is an attribute error
```

datadavev (approved):

Issue #118, opened 2022-01-12
plausible.io analytics integration in API


dannymandel:
Implement some support for plausible.io analytics integration in our API

dannymandel:
This one is done

PR #119, opened 2022-01-12
Plausible.io analytics support

dannymandel:
Initial support for analytics: https://github.com/isamplesorg/isamples_inabox/issues/118

plausible.io is running in Docker locally here.  I've configured according to their self-hosted instructions: https://plausible.io/docs/self-hosting

We have one site configured, isamples.org.

### DIFF
--- a/isb_web/analytics.py
+++ b/isb_web/analytics.py
@@ -1,5 +1,7 @@
 import logging
 import typing
+
+import fastapi
 import requests
 import json
 import isb_web.config
@@ -29,18 +31,14 @@ class AnalyticsEvent(_NoValue):
 
 def record_analytics_event(
     event: AnalyticsEvent,
-    caller_user_agent: str,
-    caller_ip: str,
-    caller_url: str,
+    request: fastapi.Request,
     properties: typing.Optional[typing.Dict] = None,
 ) -> bool:
     """
     Records an analytics event in plausible.io
     Args:
         event: The event to record
-        caller_user_agent: User agent of the caller
-        caller_ip: IP of the caller
-        caller_url: URL the caller used to invoke the event
+        request: The fastapi.Request object containing all the caller info
         properties: Custom event properties
 
     Returns: true if plausible responds with a 202, false otherwise
@@ -49,13 +47,15 @@ def record_analytics_event(
     if ANALYTICS_URL == "UNSET":
         logging.error("Analytics URL is not configured.  Please check isb_web_config.env.")
         return False
-
+    logging.error(f"Request headers are {request.headers}")
+    logging.error(f"Request client is {request.client}")
+    logging.error(f"Request url is {request.url}")
     headers = {
         "Content-Type": MEDIA_JSON,
-        "User-Agent": caller_user_agent,
-        "X-Forwarded-For": caller_ip,
+        "User-Agent": request.headers.get("user-agent", "no-user-agent"),
+        "X-Forwarded-For": request.client.host or "no-client-ip",
     }
-    data_dict = {"name": event.value, "domain": ANALYTICS_DOMAIN, "url": caller_url}
+    data_dict = {"name": event.value, "domain": ANALYTICS_DOMAIN, "url": str(request.url)}
     if properties is not None:
         # plausible.io has a bug where it needs the props to be stringified when posted in the data
         # https://github.com/plausible/analytics/discussions/1570
--- a/isb_web/analytics.py
+++ b/isb_web/analytics.py
@@ -1,0 +1,63 @@
+import logging
+import typing
+import requests
+import json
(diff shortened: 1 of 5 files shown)


isb_web/analytics.py
```diff
@@ -0,0 +1,63 @@
...
+    THING_LIST_METADATA = "thing_list_metadata"
+    THING_LIST_TYPES = "thing_list_types"
+    THING_SOLR_SELECT = "thing_solr_select"
+    THING_SOLR_LUKE_INFO = "thing_solr_luke_info"
+    THING_SOLR_STREAM = "thing_solr_stream"
+    THING_BY_IDENTIFIER = "thing_by_identifier"
+    STAC_ITEM_BY_IDENTIFIER = "stac_item_by_identifier"
+    STAC_COLLECTION = "stac_collection"
+    RELATED_METADATA = "related_metadata"
+    RELATED_SOLR = "related_solr"
```
dannymandel, line 29: In theory we have one event per API call.  To get these to show up in the dashboard, they also need to be added as conversion goals in the plausible.io admin UI.

isb_web/analytics.py
```diff
@@ -0,0 +1,63 @@
...
+    Args:
+        event: The event to record
+        caller_user_agent: User agent of the caller
+        caller_ip: IP of the caller
+        caller_url: URL the caller used to invoke the event
+        properties: Custom event properties
+
+    Returns: true if plausible responds with a 202, false otherwise
+
+    """
```
dannymandel, line 46: This is all that's required to write these events to the API store.  Just a simple post with some JSON embedded in it.

isb_web/config.py
```diff
@@ -54,5 +54,11 @@ class Settings(BaseSettings):
 
     stac_collection_url_path: str = "stac_collection"
 
+    # The URL to the analytics API
+    analytics_url: str = "UNSET"
+
+    # The domain to record analytics events for, needs to be configured as a site in plausible.io
+    analytics_domain: str = "UNSET"
+
```
dannymandel, line 62: New config to tell us the URL to post the analytics data, and also which domain we should record data for.

isb_web/main.py
```diff
@@ -253,6 +254,8 @@ async def thing_list_metadata(
 
 @app.get(f"/{THING_URL_PATH}/", response_model=ThingPage)
 def thing_list(
+    user_agent: typing.Optional[str] = Header("no_user_agent"),
+    x_forwarded_for: typing.Optional[str] = Header("no_ip"),
```
dannymandel: I didn't spend too much time on this as I knew it would run differently behind nginx and I don't have that configured locally (yet), so I do expect this part to change a bit once I experiment there.

isb_web/main.py
```diff
@@ -262,7 +265,10 @@ def thing_list(
     total_records, npages, things = sqlmodel_database.read_things_summary(
         session, offset, limit, status, authority
     )
-
+    properties = {
+        "authority": authority or "None"
+    }
```
dannymandel, line 271: These custom properties allow us to segment the analytics events and get counts that way on the dashboard, e.g.

image

isb_web/main.py
```diff
@@ -262,7 +265,10 @@ def thing_list(
...
+    analytics.record_analytics_event(AnalyticsEvent.THING_LIST, user_agent, x_forwarded_for, "http://isamples.org", properties)
```
dannymandel: The hardcoded "http://isamples.org" is obviously wrong -- need to figure out the correct way to get that out of FastAPI when running behind nginx.

dannymandel: But, in theory, we'd have a line like this for every API call that we expose.

isb_web/config.py
dannymandel: @datadavev, I did tell you I think we'd need to define separate hostnames to get this working properly.  I read a bit more on their site, though, and it looks like the domain is just an identifier where we record data.  So, in theory, we could just use this config differently on each iSB instance in henry, and send it along to the API happily.

datadavev: OK, that's not too hard to deal with.

isb_web/main.py
datadavev: All seems quite straight forward.

The service URL should be available in `Request.url`, I think it has the external url that was called.

datadavev (commented):
Looks good so far. Nice and simple

dannymandel:
Thanks for taking a look, @datadavev!  `Request.url` did the trick.

dannymandel:
Overall, I think using the raw request object simplifies things a bit as well, as it has the same information all in one place.  Better to read everything out of there in one spot and add reasonable defaults in case pieces are missing.

Issue #126, opened 2022-02-10
iSamples SESAR entries out of sync with data in SESAR


dannymandel:
We did a count comparison and there are a number of SESAR entries not present in iSamples.

dannymandel:
The root cause of this was due to SESAR publishing test records.  We did a manual fix.

PR #127, opened 2022-02-10
Manual sesar load

dannymandel:
New manual load option to allow us to fix https://github.com/isamplesorg/isamples_inabox/issues/126

### DIFF
--- a/scripts/find_missing_sesar_records.py
+++ b/scripts/find_missing_sesar_records.py
@@ -1,5 +1,5 @@
-import sqlite3
-from isb_web.sqlmodel_database import SQLModelDAO
+# import sqlite3
+# from isb_web.sqlmodel_database import SQLModelDAO
 
 
 # Manual script to dump and compare records between SESAR and iSamples
@@ -70,5 +70,3 @@ if __name__ == "__main__":
     #             last_slash_index = loc.rfind("/") + 1
     #             after_last_slash = loc[last_slash_index:]
     #             sesar_file.write(f"{after_last_slash},{lastmod}\n")
-
-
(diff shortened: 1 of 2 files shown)


scripts/sesar_things.py
```diff
@@ -40,13 +41,16 @@ def countThings(session):
...
-async def _loadSesarEntries(session, max_count, start_from=None):  # noqa: C901 -- need to examine computational complexity
+async def _loadSesarEntries(session, max_count, start_from=None, manual_ids:typing.List[typing.List[str]] = None):  # noqa: C901 -- need to examine computational complexity
     L = getLogger()
     futures = []
     working = {}
-    ids = isb_lib.sesar_adapter.SESARIdentifiersSitemap(
-        max_entries=countThings(session) + max_count, date_start=start_from
-    )
+    if manual_ids is not None:
+        ids = iter(manual_ids)
```
dannymandel, line 49: Just manually use the ids if we have them, as opposed to building the list from the sitemap

scripts/find_missing_sesar_records.py
```diff
@@ -0,0 +1,74 @@
...
+
+    with open(sqlite_dump_file, "r") as sqlite_file:
+        for line in sqlite_file:
+            row = line.split(",")
+            sqlite_igsns[row[0]] = row[1]
+
+    linenum = 0
+    with open(isamples_dump_file, "r") as isamples_file:
+        for line in isamples_file:
+            row = line.split(",")
```
datadavev, line 21: fyi - python has an excellent built in [csv module](https://docs.python.org/3/library/csv.html) that handles things like quoted values etc.

datadavev (approved):
only suggestion is to use the csv module for parsing to minimize edge cases. Not a problem here, just fyi.

PR #130, opened 2022-02-15
Try to thin out dependencies

### LINKED ISSUE #129: Thin out python dependencies
In the course of development it seems like our python environment has gotten a bit bloated, we should see how many of the things we are actually using and attempt to get rid of anything else.

dannymandel:
https://github.com/isamplesorg/isamples_inabox/issues/129


dannymandel:
See also https://github.com/isamplesorg/igsn_lib/pull/3

PR #132, opened 2022-02-23
Changes for new date field to allow for range faceting

dannymandel:
@datadavev @qgan7125, these are the solr changes for the new field type and the manual script to reindex the data with the new field

### DIFF
--- a/isb_lib/core.py
+++ b/isb_lib/core.py
@@ -243,7 +243,9 @@ def handle_produced_by_fields(coreMetadata: typing.Dict, doc: typing.Dict):  # n
         raw_date_str = producedBy["resultTime"]
         date_time = parsed_date(raw_date_str)
         if date_time is not None:
-            doc["producedBy_resultTime"] = datetimeToSolrStr(date_time)
+            solr_date_str = datetimeToSolrStr(date_time)
+            doc["producedBy_resultTime"] = solr_date_str
+            doc["producedBy_resultTimeRange"] = solr_date_str
     if _shouldAddMetadataValueToSolrDoc(producedBy, "@id"):
         doc["producedBy_isb_core_id"] = producedBy["@id"]
     if "samplingSite" in producedBy:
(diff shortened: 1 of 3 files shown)


isb_lib/core.py
```diff
@@ -243,7 +243,9 @@ def handle_produced_by_fields(coreMetadata: typing.Dict, doc: typing.Dict):  # n
         raw_date_str = producedBy["resultTime"]
         date_time = parsed_date(raw_date_str)
         if date_time is not None:
-            doc["producedBy_resultTime"] = datetimeToSolrStr(date_time)
+            solr_date_str = datetimeToSolrStr(date_time)
+            doc["producedBy_resultTime"] = solr_date_str
+            doc["producedBy_resultTimeRange"] = solr_date_str
```
dannymandel, line 248: Currently we are populating two date fields -- I think we should be able to get rid of `producedBy_resultTime`, though.

scripts/solr_reindex_template.py
```diff
@@ -11,7 +11,9 @@
 
 def mutate_record(record: typing.Dict) -> typing.Optional[typing.Dict]:
     # Do whatever work is required to mutate the record to update things…
-    return None
+    record_copy = record.copy()
+    record_copy["producedBy_resultTimeRange"] = record.get("producedBy_resultTime")
```
dannymandel, line 15: This one line adds the new field into the record during reindex

solr_schema_init/create_isb_core_schema.py
```diff
@@ -216,4 +216,11 @@ def createCollection():
 createField("producedBy_samplingSite_location_latitude", "pfloat", True, True, None)
 createField("producedBy_samplingSite_location_longitude", "pfloat", True, True, None)
```
dannymandel, line 218: New field type and field.  Interestingly `DateRangeField` isn't included as part of the default field types in Solr (though it is part of the distribution).

PR #133, opened 2022-02-23
Bump astropy dependency as the one we had was failing to deploy on henry

dannymandel:
@datadavev FYI -- the version we previously had suddenly claimed it couldn't be found when attempting to deploy the latest version of things on Henry.  Bumping to 5.0 fixed the issue.


dannymandel:
See also https://github.com/isamplesorg/igsn_lib/commit/e9adc8903451c4bba356592a7f90748c9ed04f6a

The GitHub action failed on that commit, not sure why.

dannymandel:
This one looks bad, too, though the version definitely works on Henry, as that is the one I tested there to get the deploy to actually succeed.  I'll need to investigate a bit more

dannymandel:
OK, mystery solved -- we can't update the `pyproject.toml` without also updating `poetry.lock`, or things get out of whack.

PR #140, opened 2022-03-22
Improve consume sitemaps

dannymandel:
On inspection of the sitemap consumption code, there were a few low-hanging fruit that improved performance:

- Perhaps most importantly, we don't wait for the futures to completely drain before fetching new ones
- Turn down the logging, as it was pretty noisy and that impacted performance a bit
- Up the max futures in flight to 50 from 20
- Fixed a bug where we weren't properly utilizing the connection pool max size, which caused recreation of the pool

Things I tried that didn't help:

- Adding an in-memory dirty check that we tested before running the db save (actually made things slower)

Things I didn't try that might be helpful:

- Building a cache of known existing Thing IDs -- currently we try and save and catch the integrity exception, probably isn't as efficient.
- Implementing a proper batch things API, so rather than fetching 50 things with 50 requests -- fetch 50 things with 1 request.

All in all it was a noticeable improvement, we got things down to ~37 seconds for 1000 records, down from a minute or so.

### DIFF
--- a/scripts/consume_sitemaps.py
+++ b/scripts/consume_sitemaps.py
@@ -19,7 +19,7 @@ from isb_lib.sitemaps.sitemap_fetcher import (
 from isb_web import sqlmodel_database
 from isb_web.sqlmodel_database import SQLModelDAO
 
-CONCURRENT_DOWNLOADS = 20
+CONCURRENT_DOWNLOADS = 50
 # when we hit this length, add some more to the queue
 REFETCH_LENGTH = CONCURRENT_DOWNLOADS / 2
 
--- a/scripts/consume_sitemaps.py
+++ b/scripts/consume_sitemaps.py
@@ -1,5 +1,7 @@
 import concurrent.futures
 import urllib.parse
+from concurrent.futures import ThreadPoolExecutor
+from typing import Iterator
 
 import click
 import requests
@@ -18,6 +20,8 @@ from isb_web import sqlmodel_database
 from isb_web.sqlmodel_database import SQLModelDAO
 
 CONCURRENT_DOWNLOADS = 20
+# when we hit this length, add some more to the queue
+REFETCH_LENGTH = CONCURRENT_DOWNLOADS / 2
 
 
 @click.command()
@@ -49,11 +53,12 @@ def main(ctx, url: str, authority: str, ignore_last_modified: bool):
         pool_connections=CONCURRENT_DOWNLOADS * 2, pool_maxsize=CONCURRENT_DOWNLOADS * 2
     )
     rsession.mount("http://", adapter)
+    rsession.mount("https://", adapter)
     db_url = isb_web.config.Settings().database_url
     db_session = SQLModelDAO(db_url).get_session()
     if authority is not None:
         authority = authority.upper()
-    isb_lib.core.things_main(ctx, db_url, solr_url, "INFO", False)
+    isb_lib.core.things_main(ctx, db_url, solr_url, "CRITICAL", False)
     if ignore_last_modified:
         last_updated_date = None
     else:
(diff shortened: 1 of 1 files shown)


scripts/consume_sitemaps.py
```diff
@@ -17,7 +19,9 @@
 from isb_web import sqlmodel_database
 from isb_web.sqlmodel_database import SQLModelDAO
 
-CONCURRENT_DOWNLOADS = 20
+CONCURRENT_DOWNLOADS = 50
```
dannymandel, line 22: Bump the max requests up

scripts/consume_sitemaps.py
```diff
@@ -17,7 +19,9 @@
...
+# when we hit this length, add some more to the queue
+REFETCH_LENGTH = CONCURRENT_DOWNLOADS / 2
```
dannymandel, line 24: Introduce a new variable to kick off prefetching of new records while the current ones are being processed.

scripts/consume_sitemaps.py
```diff
@@ -49,11 +53,12 @@ def main(ctx, url: str, authority: str, ignore_last_modified: bool):
         pool_connections=CONCURRENT_DOWNLOADS * 2, pool_maxsize=CONCURRENT_DOWNLOADS * 2
     )
     rsession.mount("http://", adapter)
+    rsession.mount("https://", adapter)
```
dannymandel, line 56: This was a bug that prevented us from using the increased connection pool size.

scripts/consume_sitemaps.py
```diff
@@ -49,11 +53,12 @@ def main(ctx, url: str, authority: str, ignore_last_modified: bool):
...
     db_url = isb_web.config.Settings().database_url
     db_session = SQLModelDAO(db_url).get_session()
     if authority is not None:
         authority = authority.upper()
-    isb_lib.core.things_main(ctx, db_url, solr_url, "INFO", False)
+    isb_lib.core.things_main(ctx, db_url, solr_url, "CRITICAL", False)
```
dannymandel, line 61: Turn down the logging as the chattiness was costing a few seconds per thousand records.

scripts/consume_sitemaps.py
```diff
@@ -79,6 +86,26 @@ def thing_fetcher_for_url(thing_url: str, rsession) -> ThingFetcher:
...
+    while len(thing_futures) < CONCURRENT_DOWNLOADS:
+        try:
+            thing_fetcher = thing_fetcher_for_url(next(sitemap_file_iterator), rsession)
+            # Add the download job to the queue and remember it
+            thing_future = thing_executor.submit(thing_fetcher.fetch_thing)
+            thing_futures.append(thing_future)
+        except StopIteration:
+            logging.info(f"Finished all the requests for f{sitemap_file_fetcher.url}")
+            fetched_all_things_for_current_sitemap_file = True
+    return fetched_all_things_for_current_sitemap_file
```
dannymandel, line 106: Refactor the future construction to a method as we now call this in multiple places.

scripts/consume_sitemaps.py
```diff
@@ -135,8 +158,21 @@ def fetch_sitemap_files(authority, last_updated_date, rsession, url, db_session)
                     else:
                         logging.error(f"Error fetching thing for {thing_fetcher.url}")
                         thing_id = thing_fetcher.thing_identifier()
-                        sqlmodel_database.mark_thing_not_found(db_session, thing_id, thing_fetcher.url)
+                        sqlmodel_database.mark_thing_not_found(
+                            db_session, thing_id, thing_fetcher.url
+                        )
                     thing_futures.remove(thing_fut)
+                    if len(thing_futures) < REFETCH_LENGTH:
```
dannymandel, line 165: New future construction site to prefetch futures while we work on saving the other ones to the db.

datadavev (approved):

PR #141, opened 2022-03-23
Sqlite dump tool

### LINKED ISSUE #78: Generate an sqlite dump of records
sqlite is a single file relational database that provides a convenient mechanism for sharing data. 

Prototype generation of an sqlite database that contains the isample core record representation. Schema for the database may initially follow the things structure, though will likely be more normalized for efficiency.

dannymandel:
As requested by @hongcui, we'd like an easy to consume dump of our ISB Core records.  We already had https://github.com/isamplesorg/isamples_inabox/issues/78, so it seemed like a good time to take care of two things at once.

Implementation ends up being pretty straightforward -- create a SQLModel and use each row of solr results to create a new model object then save it with SQLAlchemy.

### DIFF
--- a/isb_lib/models/isb_core_record.py
+++ b/isb_lib/models/isb_core_record.py
@@ -8,184 +8,248 @@ class ISBCoreRecord(SQLModel, table=True):
         default=None,
         nullable=False,
         primary_key=True,
-        description="The identifier in the original data, which should be a unique identifier (e.g. ark or igsn)"
+        description="The identifier in the original data, which should be a unique identifier (e.g. ark or igsn)",
     )
     isb_core_id: Optional[str] = Field(
         default=None,
         nullable=False,
-        description="The iSamples persistent identifier, which is how the sample is uniquely identified in iSamples"
+        description="The iSamples persistent identifier, which is how the sample is uniquely identified in iSamples",
     )
     isb_updated_time: Optional[str] = Field(
         default=None,
         nullable=False,
         description="The time the record was last updated in the iSamples index",
-        index=False
+        index=False,
     )
-    source_updated_time: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The time the record was last updated in the original data, if available",
-        index=False
-    ),
-    label: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A label for the record",
-        index=False
-    ),
-    search_text: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A collection of text suitable for use in a full-text search index, semicolon delimited",
-        index=False
-    ),
-    sample_description: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A textual description of the sample",
-        index=False
-    ),
-    context_categories: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The context categories from the iSamples controlled vocabulary, semicolon delimited",
-        index=False
-    ),
-    material_categories: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The material categories from the iSamples controlled vocabulary, semicolon delimited",
-        index=False
-    ),
-    specimen_categories: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The material categories from the iSamples controlled vocabulary, semicolon delimited",
-        index=False
-    ),
-    keywords: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="Keywords for searching, semicolon delimited",
-        index=False
-    ),
-    informal_classification: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="Informal scientific classification of the sample",
-        index=False
-    ),
-    produced_by_isb_core_id: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The iSB Core ID of the sample that produced this sample",
-        index=False
-    ),
-    produced_by_label: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A label for how the sample was produced",
-        index=False
-    ),
-    produced_by_description: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A text description for how the sample was produced",
-        index=False
-    ),
-    produced_by_feature_of_interest: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="A string specifying whether the sample was produced by a feature of interest",
-        index=False
-    ),
-    produced_by_responsibility: Optional[str] = Field(
-        default=None,
-        nullable=True,
-        description="The name of the people or institution responsible for producing the sample, semicolon-delimited",
-        index=False
-    ),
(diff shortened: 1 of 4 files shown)


isb_lib/models/isb_core_record.py
```diff
@@ -0,0 +1,255 @@
+from typing import Optional
+
+from sqlmodel import SQLModel, Field
+
+
```
dannymandel, line 5: This ends up being pretty much as you'd expect -- one field == one column in the SQLite schema.

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+import isb_lib.core
+from isamples_metadata import Transformer
+from isb_lib.models.isb_core_record import ISBCoreRecord
+from isb_web.isb_solr_query import ISBCoreSolrRecordIterator
+import requests
+from sqlmodel import SQLModel, create_engine, Session
+import logging
+
+# Use this to map between the solr column names and the sqlite column names and control what is
+# written to the dump.
```
dannymandel, line 15: Here we declare which fields we want to include in the dump, as well as rename them when appropriate.

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+    "curation_label": "curation_label",
+    "curation_description": "curation_description",
+    "curation_accessConstraints": "curation_access_constraints",
+    "curation_location": "curation_location",
+    "curation_responsibility": "curation_responsibility",
+    "relatedResource_isb_core_id": "related_resources_isb_core_id",
+    "source": "source"
+}
+
+NUMERIC_COLUMNS = ["produced_by_sampling_site_elevation_in_meters"]
```
dannymandel: This ended up being the only real non-string column.  Everything else made sense as a string in SQLite.

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+# Filter out any placeholder values that shouldn't appear in the dump
+def _filtered_value(value):
+    if type(value) is str and value == Transformer.NOT_PROVIDED:
```
dannymandel, line 61: On inspection of the data, we had a fair number of "Not Provided" in the solr data.  Filter these out so they don't end up in the sqlite dump.

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+    "--authority",
+    default=None,
+    help="Which authority to use when selecting and dumping the resolved_content",
+)
+@click_config_file.configuration_option(config_file_name="isb.cfg")
+@click.pass_context
+def main(ctx, db_url, solr_url, verbosity, heart_rate, authority):
+    isb_lib.core.things_main(ctx, db_url, solr_url, verbosity, heart_rate)
+    engine = create_engine(ctx.obj["db_url"], echo=False)
+    SQLModel.metadata.drop_all(engine, tables=[ISBCoreRecord.__table__])
```
dannymandel, line 95: Drop the table every time…made sense for how we're using this and certainly made development easier.

dannymandel: Also, explicitly specify the table so we don't pull in any of the other stuff (Things and related indexes).

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+    rsession = requests.session()
+    iterator = ISBCoreSolrRecordIterator(rsession, authority, batch_size, 0, "id asc")
+    num_records = 0
+    for solr_record in iterator:
+        new_record = ISBCoreRecord()
+        for key, value in SOLR_TO_SQLITE_FIELD_MAPPINGS.items():
+            solr_value = solr_record.get(key)
+            if type(solr_value) is list:
+                filtered_solr_values = [_filtered_value(value) for value in solr_value]
+                solr_value = "; ".join(filtered_solr_values)
```
dannymandel: For solr lists, do two things:

(1) Remove any embedded "Not Provided" strings
(2) Turn the list into a semicolon separated string.  SQLite doesn't have a list data type.  Not going to worry about trying to do anything goofy like make a join table for these.

@datadavev if you favor a different approach let me know

scripts/create_sql_lite_dump.py
```diff
@@ -0,0 +1,125 @@
...
+            elif type(solr_value) is str:
+                # Don't include placeholder empty values in the dump
+                solr_value = _filtered_value(solr_value)
+            # Key is source column solr name, value is dest column sqlite name in sqlite
+            if solr_value is not None:
+                new_record.__setattr__(value, solr_value)
+            else:
```
dannymandel, line 124: This last bit is a workaround for a hack somewhere below me in the stack (not sure if SQLModel or SQLAlchemy, or SQLite) -- basically if we didn't explicitly initialize a value for a column, it ended up using the column metadata definition as the value and the generated SQL exploded.  I didn't do work to try and pinpoint how/why but verified this workaround works.

scripts/create_sql_lite_dump.py
datadavev: Aren't the various `_categories` fields multi-valued? Should these be strings or a list in the resulting ssqlite?

scripts/create_sql_lite_dump.py
datadavev: so "Not Provided" becomes a null in sqlite?

scripts/create_sql_lite_dump.py
datadavev: Replace "Not Provided" with nulls / empty. 

Wrt multivalued fields, sqlite does have JSON support, so could use a JSON field type with a list. Semi-colon delimited is a good option also, but perhaps start with a JSON list and see what sorts of barriers that raises wrt to consumption. If we can do lists, then that opens a lot of flexibility.

scripts/create_sql_lite_dump.py
datadavev: Make the column nullable? Then it should accept creation without a value

datadavev (commented):
A couple suggestions, but otherwise looks good

scripts/create_sql_lite_dump.py
dannymandel: It becomes an empty string due to the other weirdness I ran into, but yes, we don't include "Not Provided" in the dump.

scripts/create_sql_lite_dump.py
dannymandel: Yep -- Not Provided is filtered out.  

The multivalued fields are now stored as JSON lists.  The raw column in the table is a string, but SQLAlchemy takes care of converting it to a list as the record is retrieved.

scripts/create_sql_lite_dump.py
dannymandel: Has nothing to do with the column being nullable, unfortunately, I think it's a bug elsewhere in the stack.  Interestingly, the bug persisted even after switching over to the SQLAlchemy Core API.

PR #143, opened 2022-04-05
Issue 142

datadavev:
Adds support for duplicate parameters (e.g. multiple `fq=` params) in requests to the streaming API


dannymandel:
The code looks good to me -- as I mentioned on the call the only question I have is whether we really need all these different lat/lon fields.  I think the `rpt` ones were for a different library.  If we don't need both it would be good to eliminate them.  Also it looks like the unit tests failed.

dannymandel (approved):
LGTM when we get the unit tests green.

datadavev:
Interestingly, the first two runs of the test harness were errors in the test harness itself pulling dependencies.

PR #145, opened 2022-04-13
Perf tuning import 2

dannymandel:
Second round of changes from our import tuning work


dannymandel:
@datadavev going to push this out to Hyde and kick off the full imports there

isb_lib/core.py
```diff
@@ -299,8 +299,8 @@ def parsed_datetime_from_isamples_format(raw_date_str) -> datetime.datetime:
     time = time.replace("Z", "")
     time_pieces = time.split(":")
     lastmod_date = datetime.datetime(year=int(date_pieces[0]), month=int(date_pieces[1]), day=int(date_pieces[2]),
-                                     hour=int(time_pieces[0]), minute=int(time_pieces[1]), second=int(time_pieces[2]),
-                                     tzinfo=None)
+                                     hour=int(time_pieces[0]), minute=int(time_pieces[1]),
+                                     second=int(float(time_pieces[2])), tzinfo=None)
```
dannymandel, line 303: And unrelated fix I ran into when testing Smithsonian -- those timestamps had milliseconds in the seconds, which didn't parse properly as ints.  I think we never hit it because we haven't updated the Smithsonian records.

isb_lib/geome_adapter.py
```diff
@@ -119,7 +119,7 @@ def recordsInProject(self, project_id, record_type):
                 expedition_modified_datetime = dateparser.parse(expedition_modified_date)
                 # For each expedition, check to see if the mod date is greater than the last time we fetched.
                 # If it is, fetch the bcids of the samples in the expedition.
-                if expedition_modified_datetime.timestamp() > self._date_start.timestamp():
+                if self._date_start is None or expedition_modified_datetime.timestamp() > self._date_start.timestamp():
```
dannymandel, line 122: Another testing bug -- if we weren't doing a partial update we were crashing -- hit during manual testing on a full record fetch.

PR #148, opened 2022-04-15
Workaround duplicate insert bug

dannymandel:
@datadavev this is my best guess as to why that integrity constraint violation was occurring.  I can't reproduce locally but I think it's because my machine is too fast, to reproduce I think we'd need the futures to bump into one another


dannymandel:
It seems like the right code either way -- I might try and insert some sleeps to get my machine to behave slowly like cyverse

dannymandel:
The key was that the unique constraint that failed appeared to be the first thing in the batch insert -- like it was failing because it was reinserting the next time around.

dannymandel:
It does seem like this fixed it.

PR #149, opened 2022-04-19
Update opencontext

dannymandel:
@datadavev this was what I mentioned this morning -- this is the last piece to move OpenContext URIs over to citation URI (and match the previous work we implemented to use these IDs)

### DIFF
--- a/isb_lib/opencontext_adapter.py
+++ b/isb_lib/opencontext_adapter.py
@@ -175,6 +175,18 @@ def reparse_as_core_record(thing: isb_lib.models.thing.Thing) -> typing.List[typ
         raise
 
 
+def identifier_from_thing_dict(thing_dict: typing.Dict) -> str:
+    opencontext_id = thing_dict["citation uri"]
+    # check the type here because some records don't have it and have a boolean False instead
+    if opencontext_id is not None and type(opencontext_id) is str:
+        opencontext_id = isb_lib.normalized_id(opencontext_id)
+    else:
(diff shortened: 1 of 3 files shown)


isb_lib/opencontext_adapter.py
```diff
@@ -175,6 +175,18 @@ def reparse_as_core_record(thing: isb_lib.models.thing.Thing) -> typing.List[typ
         raise
 
 
+def identifier_from_thing_dict(thing_dict: typing.Dict) -> str:
+    opencontext_id = thing_dict["citation uri"]
+    # check the type here because some records don't have it and have a boolean False instead
+    if opencontext_id is not None and type(opencontext_id) is str:
+        opencontext_id = isb_lib.normalized_id(opencontext_id)
+    else:
+        opencontext_id = thing_dict["uri"]
```
dannymandel, line 184: I really don't know what else to do here -- we are going to end up in a situation where the things live on the box but don't exist in solr or in iSC.  However, given the way things are set up, we do need this identifier to proceed.

scripts/consume_sitemaps.py
```diff
@@ -252,6 +250,9 @@ def fetch_sitemap_files(
                     logging.info(
                         f"About to process {len(things_fetcher.json_things)} things"
                     )
+                    current_existing_things_batch = []
+                    current_new_things_batch = []
+
```
dannymandel, line 255: This was just some cleanup I had implemented last week and forgot to check in

PR #155, opened 2022-05-03
Smithsonian updates

dannymandel:
Changes to ingest newest Smithsonian DWC-A using the sqlalchemy Core API


dannymandel:
@datadavev -- converted to use the Core API -- takes ~1 minute to import the whole thing.

PR #157, opened 2022-05-12
Datacite exploration

dannymandel:
Initial Datacite wrapped functionality

### DIFF
--- a/isb_lib/datacite/__init__.py
+++ b/isb_lib/datacite/__init__.py
@@ -2,12 +2,49 @@ import json
 import typing
 import logging
 import requests
+from requests import Response
 from requests.auth import HTTPBasicAuth
-
-import isb_web.config
 
 CONTENT_TYPE = "application/vnd.api+json"
 DATACITE_URL = "https://api.test.datacite.org"
+
+
+def _dois_url():
+    dois_url = f"{DATACITE_URL}/dois"
+    return dois_url
+
+
+def _dois_headers():
+    headers = {"content-type": CONTENT_TYPE}
+    return headers
+
+
+def _dois_auth(password, username):
+    auth = HTTPBasicAuth(username, password)
+    return auth
+
+
+def _validate_response(response) -> bool:
+    if response.status_code < 200 or response.status_code >= 300:
+        logging.error(
+            "Error requesting new DOI, status code: %s, response %s",
+            response.status_code,
+            str(response.json()),
+        )
+        return False
+    return True
+
+
+def _post_to_datacite(
+    rsession: requests.session, post_data_str: str, username: str, password: str
+) -> Response:
+    response = rsession.post(
+        _dois_url(),
+        headers=_dois_headers(),
+        data=post_data_str,
+        auth=_dois_auth(password, username),
+    )
+    return response
 
 
 def create_draft_doi(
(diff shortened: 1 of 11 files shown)


isb_lib/datacite/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+    json_response = response.json()
+    draft_id = json_response["data"]["id"]
+    # use the DOI prefix since we creating DOIs with datacite
+    return doi_from_id(draft_id)
+
+
+# See datacite docs here: https://support.datacite.org/docs/api-create-dois
+def datacite_metadata_from_core_record(
+    prefix: str, doi: str, authority: str, core_record: dict
+) -> dict:
```
dannymandel: @datadavev I'd really appreciate your eyes on this part.  This is the skeleton of the mapping that we register with datacite.

isb_lib/identifiers/identifier.py
```diff
@@ -0,0 +1,68 @@
...
+        self._titles = titles
+        self._publisher = publisher
+        self._publication_year = publication_year
+        self._resource_type = resource_type
+
+    def metadata_dict(self) -> dict[str: Any]:
+        pass
+
+
+class DataciteIdentifier(Identifier):
```
datadavev, line 32: Could add a __str__ method that presents the instance as an identifier string, e.g. `doi:10.1234/blah` and override in subclass, e.g. `igsn:10.1234/foo`

datadavev (approved):
Looks good. I think this is sufficient for now, providing the basic functionality that's needed.

Actually - one important property that we need to add is the target of the identifier (i.e. the `url` property in DataCite) to create ["findable" identifiers](https://support.datacite.org/docs/api-create-dois#create-a-findable-doi).

It's a bit of a chicken - egg sort of deal, where the url to the item should include the identifier, but we don't know the identifier until it's created (unless the value is provided when creating).

isb_lib/identifiers/identifier.py
dannymandel: Good idea.

dannymandel:
re: The URLs, I think something comes back in the response when we create actual non-draft identifiers.  I'll double check and make any necessary changes.

dannymandel:
OK, so even if we create non-draft identifiers (with required metadata), they don't have URLs associated with them:

```
"url":null,"contentUrl":null
```

dannymandel:
I think the URL is just another piece of metadata we can associate with the identifier.  I'll add it to the class, but I think that is effectively the iSamples URL that we'd use to look up the piece of content, not something that DataCite assigns.

dannymandel:
And, on that link in the docs, they mention the required fields to make something "findable", which we already populate:

```
(DOI, creators, title, publisher, publicationYear, resourceTypeGeneral)
```

I do think it makes sense to add an action to the management script to publish an identifier.  I can work on that today.

datadavev:
> I think the URL is just another piece of metadata we can associate with the identifier. I'll add it to the class, but I think that is effectively the iSamples URL that we'd use to look up the piece of content, not something that DataCite assigns.

Yes, correct. The owner of the identifier is responsible for setting the metadata properties, including the URL to which the identifier resolves. The URL should point to the "source of truth" for the identifier, which depends a bit on the context of the user. I some cases this may be isamples central, others the iSB instance, or perhaps even the original collection (if not stored directly in iSB). For now, it's probably good to point to iSC and ensure that it will be able to redirect if necessary to other sources.

dannymandel:
Merging this now.  Will make a separate PR for the publication functionality.

PR #158, opened 2022-05-18
Mypy

dannymandel:


dannymandel:
@datadavev -- This is a first go at getting `mypy` hooked up as part of our development process.  Right now, we have a strict inclusions list.  Eventually I'd like to turn that into an exclusions list.  There is a config file detailing which checks we want to run.

.github/workflows/python-app.yml
```diff
@@ -62,4 +50,9 @@
...
        run:  |
          source ./.venv/bin/activate
-         python -m flake8 --count --max-complexity 10 --ignore E501,W503 --show-source --statistics --exclude ./.venv .
\ No newline at end of file
+         python -m flake8 --count --max-complexity 10 --ignore E501,W503 --show-source --statistics --exclude ./.venv .
+     - name: Run mypy
+       run: |
+         source ./.venv/bin/activate       
+         pip install mypy
+         mypy --install-types --non-interactive --namespace-packages
```
dannymandel, line 58: This is all we're doing to run it.

isb_lib/identifiers/datacite/__init__.py
```diff
@@ -35,7 +35,7 @@ def _validate_response(response) -> bool:
 
 
 def _post_to_datacite(
-    rsession: requests.session, post_data_str: typing.AnyStr, username: str, password: str
+    rsession: requests.Session, post_data_str: typing.AnyStr, username: str, password: str
```
dannymandel, line 38: You'll see this same error fixed in a number of different places.  There is no such type as `requests.session` 🤦‍♂️

isb_lib/models/isb_core_record.py
```diff
@@ -21,21 +21,17 @@ class ISBCoreRecord(SQLModel, table=True):
...
-            default=None,
-            nullable=True,
-            description="The time the record was last updated in the original data, if available",
-            index=False,
-        ),
+    source_updated_time: Optional[str] = Field(
+        default=None,
+        nullable=True,
+        description="The time the record was last updated in the original data, if available",
+        index=False,
```
dannymandel, line 28: Turns out I had introduced a number of compiler errors when writing this file.  Somehow it still ran and compiled successfully, but the types were all wrong.

isb_lib/sitemaps/sitemap_fetcher.py
```diff
@@ -1,8 +1,9 @@
+from __future__ import annotations
```
dannymandel, line 1: This is the right way to have forward class declarations in python!

isb_lib/sitemaps/sitemap_fetcher.py
```diff
@@ -13,14 +14,6 @@
 import json
```
dannymandel, line 16: These broken forward class declarations broke mypy -- using the annotations fixed it.

isb_lib/sitemaps/sitemap_fetcher.py
```diff
@@ -94,7 +87,7 @@ def fetch_thing(self) -> ThingFetcher:
             # self.thing = None
             return self
 
-    def thing_identifier(self) -> str:
+    def thing_identifier(self) -> Optional[str]:
```
dannymandel, line 90: Saying something was a `str` and returning `None` was a common error I had made.  Changing to `Optional` is the right fix as we legit return `None` in error cases, so the caller should check it.

isb_lib/sitemaps/thing_sitemap.py
```diff
@@ -31,15 +33,13 @@ def __init__(
         max_length: int,
         things: typing.List[typing.Dict[typing.AnyStr, typing.AnyStr]],
     ):
-        self._things = things
+        self._things: list = things
```
dannymandel, line 36: I didn't know you could do this with iVars, but this is the way to do it according to the mypy reference.

mypy.ini
```diff
@@ -0,0 +1,9 @@
+[mypy]
+python_version = 3.9
+
+files = isb_lib/identifiers,isb_lib/models,isb_lib/sitemaps
```
dannymandel: This controls which files we type check.

.github/workflows/python-app.yml
dannymandel: I tried a couple of pre-existing actions to get it to go and had no luck so just ran it manually on the command-line.  The actions supposedly annotate the PR with errors, but otherwise just running it likes this seems to do the trick just as well.

dannymandel:
I'll keep on adding things to the list, and eventually turn it into an exclusion list, but I'm happy with this first batch of files

isb_lib/sitemaps/sitemap_fetcher.py
datadavev: aha!

isb_lib/sitemaps/sitemap_fetcher.py
datadavev: I've also seen `typing.Union[str, None]`, (I believe Optional is shorthand for Union) and [python 3.10 now](https://peps.python.org/pep-0604/) supports `v: str | None` which is nice if everything is on the version

datadavev:
This is very nice. Will definitely make life simpler in the future.

isb_lib/identifiers/datacite/__init__.py
datadavev: oops

PR #159, opened 2022-05-24
Mypy part2

dannymandel:
Last batch of mypy-related fixes

### DIFF
--- a/isb_lib/geome_adapter.py
+++ b/isb_lib/geome_adapter.py
@@ -5,6 +5,8 @@ import logging
 import functools
 import datetime
 import typing
+from typing import Optional
+
 import dateparser
 
 import isamples_metadata.GEOMETransformer
@@ -321,7 +323,7 @@ def reparseThing(thing):
     return thing
 
 
-def loadThing(identifier: str, t_created: datetime.datetime, existing_thing: Thing) -> Thing:
+def loadThing(identifier: str, t_created: datetime.datetime, existing_thing: Optional[Thing]) -> Thing:
     L = getLogger()
     L.info("loadThing: %s", identifier)
     response = getGEOMEItem_json(identifier, verify=True)
@@ -363,10 +365,11 @@ def reloadThing(thing):
 def reparseAsCoreRecord(thing: Thing) -> typing.List[typing.Dict]:
     _validateResolvedContent(thing)
     core_records = []
-    transformer = isamples_metadata.GEOMETransformer.GEOMETransformer(thing.resolved_content, thing.tcreated)
-    parent_core_record = isb_lib.core.coreRecordAsSolrDoc(transformer)
-    core_records.append(parent_core_record)
-    for child_transfomer in transformer.child_transformers:
-        child_core_record = isb_lib.core.coreRecordAsSolrDoc(child_transfomer)
-        core_records.append(child_core_record)
+    if thing.resolved_content is not None:
+        transformer = isamples_metadata.GEOMETransformer.GEOMETransformer(thing.resolved_content, thing.tcreated)
+        parent_core_record = isb_lib.core.coreRecordAsSolrDoc(transformer)
+        core_records.append(parent_core_record)
+        for child_transfomer in transformer.child_transformers:
+            child_core_record = isb_lib.core.coreRecordAsSolrDoc(child_transfomer)
+            core_records.append(child_core_record)
     return core_records
--- a/isb_lib/opencontext_adapter.py
+++ b/isb_lib/opencontext_adapter.py
@@ -158,18 +158,18 @@ class OpenContextRecordIterator(isb_lib.core.IdentifierIterator):
             return
         self._page_offset = 0
 
-    def last_url_str(self) -> typing.AnyStr:
+    def last_url_str(self) -> str:
         return self.url
 
 
-def _validate_resolved_content(thing: isb_lib.models.thing.Thing):
-    isb_lib.core.validate_resolved_content(OpenContextItem.AUTHORITY_ID, thing)
+def _validate_resolved_content(thing: isb_lib.models.thing.Thing) -> dict:
+    return isb_lib.core.validate_resolved_content(OpenContextItem.AUTHORITY_ID, thing)
 
 
 def reparse_as_core_record(thing: isb_lib.models.thing.Thing) -> typing.List[typing.Dict]:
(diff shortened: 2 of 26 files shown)


isb_lib/opencontext_adapter.py
```diff
@@ -205,7 +205,9 @@ def load_thing(
     """
     L = get_logger()
     id = identifier_from_thing_dict(thing_dict)
-    t_created = dateparser.parse(thing_dict.get("updated"))
+    updated_time = thing_dict.get("updated")
+    if updated_time is not None:
+        t_created = dateparser.parse(updated_time)
```
dannymandel, line 211: Here's an example of a legit bug

isb_lib/smithsonian_adapter.py
```diff
@@ -93,13 +94,16 @@ def _validate_resolved_content(thing: Thing):
 def reparse_as_core_record(thing: Thing) -> typing.List[typing.Dict]:
     _validate_resolved_content(thing)
     try:
-        transformer = isamples_metadata.SmithsonianTransformer.SmithsonianTransformer(
-            thing.resolved_content
-        )
-        solr_doc = isb_lib.core.coreRecordAsSolrDoc(transformer)
-        # This isn't present in Smithsonian data.  Fall back to the value on Thing
-        solr_doc["sourceUpdatedTime"] = isb_lib.core.datetimeToSolrStr(thing.tstamp)
-        return [solr_doc]
```
dannymandel, line 102: Another real bug -- if we ever had a degenerate Thing record with no resolved_content this would have blown up.  Now, you could argue the Smithsonian should never have degenerate things since we're getting them from a local dump, but…

isb_web/crud.py
```diff
@@ -59,7 +59,7 @@ def getRelationsSolr(
     if len(q) == 0:
         q.append("*:*")
     headers = {"Accept": "application/json"}
-    params = {
+    params: dict = {
```
dannymandel, line 62: Things like this are a bit silly IMHO, but in certain cases it isn't smart enough to figure this out.

isb_web/schemas.py
```diff
@@ -41,31 +41,31 @@ class ThingType(pydantic.BaseModel):
     count: int
 
     class Config:
-        orm_mode: True
+        orm_mode = True
```
dannymandel, line 44: I am surprised this even compiled!

mypy.ini
```diff
@@ -1,7 +1,13 @@
 [mypy]
 python_version = 3.9
 
-files = isb_lib/identifiers,isb_lib/models,isb_lib/sitemaps,isb_lib/stac,isamples_metadata
+files = .
+exclude = (?x)(
+    .*sqlmodel_database\.py$
+    | main\.py$
+    | core\.py$
+  )
```
dannymandel, line 9: As of now, we have three files in the exclude list.

mypy.ini
```diff
@@ -12,4 +18,18 @@ ignore_missing_imports = True
 [mypy-h3.*]
 ignore_missing_imports = True
 [mypy-fasttext.*]
```
dannymandel, line 20: Anything listed here is a library without a typing package or native types in the library itself.

scripts/identifiers/manage_identifiers.py
```diff
@@ -105,7 +105,7 @@ def create_doi(ctx: Dict, file: BufferedReader, prefix: str, doi: str, igsn: boo
         # TODO: pipe in authority -- from where?
         new_thing.authority_id = "AUTHORITY"
         new_thing.resolved_url = datacite.dois_url()
-        new_thing.resolved_content = file_contents
+        new_thing.resolved_content = file_contents_dict
```
dannymandel, line 108: This was a legit bug!

scripts/migrations/populate_points_for_h3.py
```diff
@@ -28,7 +28,7 @@ def main(ctx, db_url, verbosity):
 
 
 def populate_points(session: Session):
-    distinct_h3_select = select(Thing.h3).filter(Thing.h3.isnot(None)).distinct()
+    distinct_h3_select = select(Thing.h3).filter(Thing.h3.isnot(None)).distinct()  # type: ignore
```
dannymandel, line 31: This is an example of whatever sqlmodel is doing making mypy unhappy.  Luckily they support just turning off type checking on a line by line basis.

isb_lib/geome_adapter.py
```diff
@@ -363,10 +365,11 @@ def reloadThing(thing):
 def reparseAsCoreRecord(thing: Thing) -> typing.List[typing.Dict]:
     _validateResolvedContent(thing)
     core_records = []
-    transformer = isamples_metadata.GEOMETransformer.GEOMETransformer(thing.resolved_content, thing.tcreated)
-    parent_core_record = isb_lib.core.coreRecordAsSolrDoc(transformer)
-    core_records.append(parent_core_record)
-    for child_transfomer in transformer.child_transformers:
-        child_core_record = isb_lib.core.coreRecordAsSolrDoc(child_transfomer)
-        core_records.append(child_core_record)
+    if thing.resolved_content is not None:
```
dannymandel, line 368: Again a case where we could have crashed if a Thing had missing content.

isb_lib/sesar_adapter.py
```diff
@@ -156,25 +158,32 @@ def _validateResolvedContent(thing: Thing):
...
     _validateResolvedContent(thing)
-    item = SESARItem(thing.id, thing.resolved_content)
-    if as_solr:
-        return item.solrRelations()
-    return item.asRelations()
+    if thing.id is not None and thing.resolved_content is not None:
+        item = SESARItem(thing.id, thing.resolved_content)
+        if as_solr:
+            return item.solrRelations()
+        return item.asRelations()  # type: ignore
```
dannymandel, line 165: This method appears to have disappeared.  This code isn't used anymore, I don't think.  It was the original relations solr schema that we aren't using anymore.

isb_web/schemas.py
datadavev: I'm surprised

scripts/migrations/populate_points_for_h3.py
datadavev: It is a bit odd - seems clear what the types are there.

datadavev (approved):
I'm liking this mypy a lot

PR #160, opened 2022-05-24
Integrate page view

dannymandel:
Turns out this is really simple using the Starlette TemplateResponse, the only tricky part was getting the formatting working correctly.

### DIFF
--- a/isb_web/main.py
+++ b/isb_web/main.py
@@ -11,6 +11,7 @@ import fastapi.templating
 import fastapi.middleware.cors
 import fastapi.responses
 import accept_types
+import json
 from fastapi.params import Query, Depends
 from pydantic import BaseModel
 from sqlmodel import Session
@@ -18,6 +19,7 @@ from sqlmodel import Session
 import isb_web
 import isamples_metadata.GEOMETransformer
 from isb_lib.core import MEDIA_GEO_JSON, MEDIA_JSON, MEDIA_NQUADS
+from isb_lib.models.thing import Thing
 from isb_web import sqlmodel_database, analytics
 from isb_web.analytics import AnalyticsEvent
 from isb_web import schemas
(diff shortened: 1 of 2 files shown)


isb_web/main.py
```diff
@@ -252,6 +254,24 @@ async def login():
 """
```
dannymandel, line 256: This is the new method -- @datadavev do you have a better suggestion for what the URL should be?

isb_web/main.py
```diff
@@ -485,31 +505,36 @@ async def get_thing(
...
-            )
-        elif authority_id == "OPENCONTEXT":
-            content = OpenContextTransformer(item.resolved_content).transform()
-        elif authority_id == "SMITHSONIAN":
-            content = SmithsonianTransformer(item.resolved_content).transform()
-        else:
-            raise fastapi.HTTPException(
-                status_code=400,
-                detail=f"Core format not available for authority_id: {authority_id}",
-            )
```
dannymandel, line 505: Refactor this to a method we can use in both the API and HTML.

isb_web/templates/thing.html
```diff
@@ -0,0 +1,68 @@
...
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
+    integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous" />
+  <link rel="stylesheet" href="https://unpkg.com/leaflet@1.8.0/dist/leaflet.css"
+    integrity="sha512-hoalWLoI8r4UszCkZ5kL8vayOGVae1oxXe/2A4AO6J9+580uKHDO3JdHb7NzwwzK5xr/Fs0W40kiNHxM9vyTtQ=="
+    crossorigin="" />
+  </link>
+  <script src="https://unpkg.com/leaflet@1.8.0/dist/leaflet.js"
+    integrity="sha512-BB3hKbKWOc9Ez/TAwyWxNXeoV9c1v6FIeYiBieIWkpLjauysF18NzgR1MBNBXf8/KABdlkX68nAhlwcDFLGPCQ=="
+    crossorigin=""></script>
+  <script type="application/ld+json">{{ thing_json|safe }}</script>
```
dannymandel, line 18: This one line here is the magic sauce that lets us pass the JSON into the template and render it.  The `safe` keyword tells it to pass it through raw, rather than escaping.  `thing_json` is the dictionary key we store the JSON string under in `main.py`.

isb_web/main.py
```diff
@@ -252,6 +254,24 @@ async def login():
...
+    if item is None:
+        raise fastapi.HTTPException(
+            status_code=404, detail=f"Thing not found: {identifier}"
+        )
+    content = await thing_resolved_content(identifier, item)
+    content_str = json.dumps(content)
+    return templates.TemplateResponse(
+        "thing.html", {
+            "request": request,
+            "thing_json": content_str
```
dannymandel, line 270: This `thing_json` needs to match up with the string we render in the html template.

PR #164, opened 2022-06-07
Initial api for checking orcid_id and token against a live orcid server

dannymandel:
Simple API for verifying that a token and orcid_id are valid

### DIFF
--- a/tests/test_orcid.py
+++ b/tests/test_orcid.py
@@ -3,8 +3,11 @@ import os
 import pytest
 from isb_lib.authorization import orcid
 
-orcid_id = "000-0003-2109-7692"
-token = "b035c8e7-60ee-4400-a6d3-9b02ff9ff1b0"
+
+# If manually testing, you can put these in to environment environment variables:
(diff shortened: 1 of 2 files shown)


tests/test_orcid.py
```diff
@@ -0,0 +1,30 @@
+import os
+
+import pytest
+from isb_lib.authorization import orcid
+
+orcid_id = "000-0003-2109-7692"
+token = "b035c8e7-60ee-4400-a6d3-9b02ff9ff1b0"
```
datadavev: Let's try to avoid putting tokens in GH, instead load from an env var (or cli argument, etc) and make a note about how to create it.

datadavev (approved):
Good for testing, and harmless here, but let's not put tokens in git.

qgan7125 (approved):

hyunssong (approved):

tests/test_orcid.py
dannymandel: Removed

PR #165, opened 2022-06-08
Implement code exchange for token

dannymandel:
Per discussion with @qgan7125, implement a fastapi method that receives the code from a redirect URI and calls over to orcid to turn it into a token payload.

Though this will change as we build up more of the supporting code, I tested like this:

- Created an account on https://sandbox.orcid.org/developer-tools
- Added an app using the public API, and added "http://localhost:8000/orcid_token" as a redirect URI
- Set the app id and secret using the `ORCID_CLIENT_ID` and `ORCID_CLIENT_SECRET` environment variables when running python
- Logged in to the orcid developer site, and copied the authorize request url, after selecting http://localhost:8000/orcid_token as the redirect URI
- Pasted the authorize request url into Chrome
- See token output in browser, verify Cookies are set as expected

### DIFF
--- a/isb_lib/authorization/orcid/__init__.py
+++ b/isb_lib/authorization/orcid/__init__.py
@@ -1,12 +1,22 @@
 import logging
+from typing import Optional
 
 import requests
 from requests import Response
+from isb_web import config
 
 ORCID_API_CONTENT_TYPE = "application/orcid+json"
 ORCID_API_VERSION = "2.1"
 ORCID_API_ENDPOINT = "record"
-ORCID_API_HOSTNAME = "pub.sandbox.orcid.org"
+ORCID_API_HOSTNAME = config.Settings().orcid_hostname
+ORCID_TOKEN_REDIRECT_URI = config.Settings().orcid_token_redirect_uri
+ORCID_CLIENT_ID = config.Settings().orcid_client_id
+ORCID_CLIENT_SECRET = config.Settings().orcid_client_secret
+ORCID_CODE_GRANT_TYPE = "authorization_code"
+
+
+def _orcid_token_url():
+    return f"https://{ORCID_API_HOSTNAME}/oauth/token"
 
 
 def _orcid_auth_url(orcid_id: str) -> str:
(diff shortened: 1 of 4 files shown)


isb_lib/authorization/orcid/__init__.py
```diff
@@ -1,12 +1,22 @@
...
+from isb_web import config
 
 ORCID_API_CONTENT_TYPE = "application/orcid+json"
 ORCID_API_VERSION = "2.1"
 ORCID_API_ENDPOINT = "record"
-ORCID_API_HOSTNAME = "pub.sandbox.orcid.org"
+ORCID_API_HOSTNAME = config.Settings().orcid_hostname
+ORCID_TOKEN_REDIRECT_URI = config.Settings().orcid_token_redirect_uri
+ORCID_CLIENT_ID = config.Settings().orcid_client_id
+ORCID_CLIENT_SECRET = config.Settings().orcid_client_secret
```
dannymandel, line 14: We are going to need to figure out how to store these properly in Docker so that they are available securely.  Checking them into a config file in plaintext is not the right way to do this.

dannymandel: Though, once they're in the Docker keystore, this code should continue to function as we'll just be using that to set the environment variables.

isb_web/main.py
```diff
@@ -210,6 +211,30 @@ async def login(request: fastapi.Request):
     return await _cli.authorize_redirect(request, OAUTH_REDIRECT_URL, state=state)
 
 
+@app.get("/orcid_token", include_in_schema=False)
+def get_orcid_token(request: fastapi.Request, response: fastapi.Response):
```
dannymandel, line 215: @datadavev I saw that the GitHub OAuth implementation had some referrer checks.  Not sure if we need to add those here?  If so, presumably the only allowed host should be orcid?

dannymandel:
@datadavev -- one other general question.  I saw that the previous GitHub integration used something from starlette to manage the requests.  Should I have tried to use that in lieu of making the requests manually?

isb_web/main.py
datadavev: Technically (though very unlikely) it could be used by anyone as a callback for github authentication. The referer checks are there to reduce the likelihood of abuse of the service.

This is a bit unusual because most of the work is done by a browser client that does not necessarily originate from that server.

datadavev:
> @datadavev -- one other general question. I saw that the previous GitHub integration used something from starlette to manage the requests. Should I have tried to use that in lieu of making the requests manually?

I think [authlib](https://docs.authlib.org/en/latest/index.html) likely provides implementations for much of what is needed for the oauth handshakes etc. The use of starlette (upon which fastapi extends) I think came from some examples in the docs.

isb_web/main.py
```diff
@@ -210,6 +211,30 @@ async def login(request: fastapi.Request):
...
+    Returns:
+        For now, the raw token payload, though this is likely to change to a ReactJS-backed page of some sort.
+    """
+    code = request.query_params.get("code")
+    token_payload = orcid.exchange_code_for_token(code, requests.session())
+    if token_payload is not None:
+        expires = token_payload.get("expires_in")
+        response.set_cookie(key="access_token", value=token_payload.get("access_token"), expires=expires)
+        response.set_cookie(key="refresh_token", value=token_payload.get("refresh_token"), expires=expires)
+        response.set_cookie(key="orcid", value=token_payload.get("orcid"), expires=expires)
```
qgan7125, line 232: Should the client-side also need to set cookies?

qgan7125 (approved):

isb_web/main.py
dannymandel: No, I think this is all we need.  I think what we'll end up doing is having this response redirect to a view page you will write.

PR #166, opened 2022-06-14
Mint identifier api

dannymandel:
See also: https://github.com/isamplesorg/isamples_docker/pull/23

Sample code that invokes it:

```
import json
import requests
json_dict = {
  "token": "337f54d4-f013-4737-a629-3544d8f5174d",
  "orcid_id": "0000-0003-2109-7692",
  "datacite_metadata": {
    "data": {
      "type": "dois",
      "attributes": {
        "prefix": "10.82273",
        "types": {
          "resourceTypeGeneral": "PhysicalObject"
        },
        "creators": [
          "Curator Integrated Ocean Drilling Program (TAMU)"
        ],
        "titles": [
          {
            "title": "Sample 178-1098B-1H-1 (143-144 cm.)"
          }
        ],
        "publisher": "AUTHORITY",
        "publicationYear": 2006
      }
    }
  }
}
post_data_str = json.dumps(json_dict).encode("utf-8")
response = requests.post("http://localhost:8000/mint_identifier", data=post_data_str)
print(f"response is {response.content}")
```

### DIFF
--- a/isb_lib/identifiers/datacite/__init__.py
+++ b/isb_lib/identifiers/datacite/__init__.py
@@ -25,10 +25,10 @@ def _dois_auth(password, username):
 
 def _validate_response(response) -> bool:
     if response.status_code < 200 or response.status_code >= 300:
+
         logging.error(
-            "Error requesting new DOI, status code: %s, response %s",
+            "Error requesting new DOI, status code: %s",
             response.status_code,
-            str(response.json()),
         )
         return False
     return True
--- a/isb_web/config.py
+++ b/isb_web/config.py
@@ -63,14 +63,20 @@ class Settings(BaseSettings):
     # The URL to the datacite API
     datacite_url: str = "https://api.test.datacite.org/"
 
+    # This shouldn't be checked in.  Set by doing export DATACITE_USERNAME=foobar
+    datacite_username: str = ""
+
+    # This shouldn't be checked in.  Set by doing export DATACITE_PASSWORD=foobar
+    datacite_password: str = ""
+
     orcid_hostname: str = "pub.sandbox.orcid.org"
 
     orcid_token_redirect_uri: str = "http://localhost:8000/orcid_token"
 
-    # This shouldn't be checked in.  Set it by doing export export ORCID_CLIENT_ID=foobar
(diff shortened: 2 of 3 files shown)


qgan7125:
```
    "type": "dois",
    "attributes": {
      "prefix": "10.82273",
      "types": {
        "resourceTypeGeneral": "PhysicalObject"
      }
```

Is this part hard code?

dannymandel:
Yes, we can hard code the `resourceTypeGeneral`.  For the prefix, we'll need to add some config to the JS app so it knows which prefix to use.  Alternately we could make it part of the server config and have the server insert the key into the JSON it receives.  That might be simpler.

dannymandel:
Going to merge this one so we can push it up to Mars.  Leave comments and I will address in subsequent PRs.

datadavev:
Should separate authentication information from the body of the request. Send the token via an `Authorization` header rather than in the body of the message.

dannymandel:
I'll work on converting the auth pieces over to headers, @datadavev.  @qgan7125 I thought a bit more about the prefix, and I think we want to include that in the JS app after all.  There are two different modes for creating identifiers -- drafts or published, and in the case of the published one, we'd likely want to enforce the prefix so that it wasn't user-editable.  

For example, the registered prefix for one of our test accounts is `10.82273`, but a fully generated identifier would be `10.82273/txrk-w112`.  We'd likely want the UI to only let the user type in the `txrk-w112` piece, as the API will break if you try and pass in a different prefix.

The draft mode would just have a checkbox and turn off most of the editable fields, or potentially would be a different UI altogether.  What do you think, @datadavev?

datadavev:
I spent some time working through the oauth2 workflow process for both browser and command line clients and put together a simple fastapi app that supports both, and would work with react etc.: https://github.com/datadavev/test_orcid_oauth

It handles JWT auth by verifying the provided token against the ORCID signing keys, and so does not need to make additional calls to ORCID to verify a token.

It may be useful as a reference, though I think there's still a lot of room for improvement.

datadavev:
re: draft mode - I think that's correct Danny - in draft many of the validation checks would not be needed.

PR #170, opened 2022-06-15
Fix GEOME h3 calculation where lat/lon were reversed

dannymandel:
Should be pretty self-explanatory, will need to re-run all the indexers manually


dannymandel:
Going to merge so I can push this up and rebuild things

PR #175, opened 2022-06-29
Further changes per discussion with Matthew

dannymandel:
Per discussion with @qgan7125 we need to redirect to the DOIs page.  Also, move this to config since we're going to need to configure differently when we get the set up on the various iSB instances.

### DIFF
--- a/isb_web/manage.py
+++ b/isb_web/manage.py
@@ -133,7 +133,6 @@ oauth.register(
 
 
 class MintIdentifierParams(BaseModel):
-    orcid_id: str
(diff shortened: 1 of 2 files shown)


isb_web/manage.py
```diff
@@ -133,7 +133,6 @@ async def __call__(
 
 
 class MintIdentifierParams(BaseModel):
-    orcid_id: str
```
dannymandel, line 136: This isn't used anymore, as it's been replaced by the user in the session.

PR #179, opened 2022-07-05
Refactor url parsing bits to a utility module to make it easy to repurpose login/logout logic

dannymandel:
Per discussion with @qgan7125 -- we need to make the logout logic go to the right page.  In order to do that, extract some of the URL parsing out to a utilities module to make this easy to reuse.


dannymandel:
I messed up this PR, going to redo it…

PR #181, opened 2022-07-07
Permissions list

dannymandel:
Have a list of orcid ids that are authorized to mint identifiers


isb_web/sqlmodel_database.py
```diff
@@ -21,7 +22,7 @@
 
 
 class SQLModelDAO:
-    def __init__(self, db_url: str, echo: bool = False):
+    def __init__(self, db_url: Optional[str], echo: bool = False):
```
dannymandel, line 25: mypy correctly pointed out that this can be `None` so it should be an `Optional`.

dannymandel:
@datadavev this is where we ended up.  Still need to restore the session if there is a JWT cookie but no session, but that's a separate task so will handle it in a separate PR.

PR #182, opened 2022-07-12
Adding conceptual documentation explaining how the authorization and identifier minting flow is expected to work.

dannymandel:
Start on some auth docs


dannymandel:
@datadavev have a look and please let me know if there are any things that you'd expect to be covered that are missing.

datadavev (approved):
Good documentation - clear and concise. Only suggestion is on generalizing from "DOI" to something like "DataCite identifiers" since both IGSN and DOI identifiers will be created using the same API (technically they are much the same, they mostly differ only from an end user perspective).

PR #183, opened 2022-07-13
Add taxonomy mapping script  -- disregard 

hyunssong:
Create a script that maps the taxonomy into json format. I was not sure what to return from this script, so I just had the main method print the created json as a string.


hyunssong:
I mistakenly pushed this directly to develop branch and created a wrong PR. please disregard!

PR #184, opened 2022-07-13
Create taxonomy mapping script

hyunssong:
Create script that maps taxonomy to json format. I was not sure what to return, so had it print the created json.


pyproject.toml
```diff
@@ -30,10 +30,11 @@ itsdangerous = "^2.1.2"
 starlette-oauth2-api = "^0.2.6"
 httpx = "^0.20.0"
 
-fasttext = "^0.9.2"
+# fasttext = "^0.9.2"
```
hyunssong, line 33: commenting this out for now, for some reason installing fasttext is causing errors and I will try to look into this.

datadavev (commented):
Good start. It may be worth making the script work on one vocabulary at a time, and specifying the vocabulary to process as a parameter to the script. That would make it much easier to generate a single JSON files for a given vocabulary. My preference for command line processing is the [click library](https://click.palletsprojects.com/en/8.1.x/). It is very easy to use for simple cases like this. Then we could run the script on which ever vocab we need, and pipe the output to a json file.

For example:
```
python create_label_hierarchy.py "https://some/vocabulary.ttl" > vocab.json
```

datadavev (approved):
Ah, I see the update version takes command line params. Perfect!

hyunssong:
I fixed the code to accept command line arguments as @datadavev suggested in `create_hierarchy_json.py` (which was orginally named `create_label_hierarchy.json`) . I added the classification code to use the model and print out the predicted label and probability in `classification.py` . For now it is iterating through records and printing out the prediction. I am not sure why I am getting pytest errors so I will look into this.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ starlette-oauth2-api = "^0.2.6"
 httpx = "^0.20.0"
 
 # fasttext = "^0.9.2"
+regex = "2022.3.2"
```
hyunssong, line 34: I was keep getting errors on pytest and it seemed like a similar issue to [this](https://github.com/domainaware/parsedmarc/issues/298) so I downgraded the version of regex.

hyunssong:
I am going to merge this into develop so Matthew can write the `README` for json generation in `isamples-webui`

PR #191, opened 2022-07-21
Taxonomy classify

hyunssong:
Based on the discussion with experts, add rules to integrate expert knowledge in our classification process. 
If the record falls into any of these rules that are defined, we will give the corresponding label.
If it does not, we pass the record to the classification model and get the machine prediction.

### DIFF
--- a/scripts/taxonomy/SESARClassifierInput.py
+++ b/scripts/taxonomy/SESARClassifierInput.py
@@ -1,0 +1,89 @@
+class SESARClassifierInput:
+    """Takes thing object to convert it into data that the classifier
+    requires as input"""
+    def __init__(self):
+        # gold label field of SESAR
+        self.SESAR_material_field = "material"
+        self.SESAR_sample_field = "sampleType"
+        # fields to fill in through parsing the thing object
+        self.description_map = {}
+        self.gold_material, self.gold_sample = None, None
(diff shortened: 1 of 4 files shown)


scripts/taxonomy/classification.py
```diff
@@ -108,21 +132,29 @@ def main(ctx, db_url: str, solr_url: str, max_records: int, verbosity: str):
...
         ).transform()
 
         # extract text and gold label from object
-        text, gold_material, gold_sample = parse_SESAR_thing(
-            thing.resolved_content
+        (
+            description_map,
+            material_text, sample_text,
+            gold_material, gold_sample
+        ) = parse_SESAR_thing(thing.resolved_content)
```
dannymandel: If these are common fields that belong together, it might make sense to refactor this to a class.  Grouped tuples are fine for work in progress but we should formalize it if we're going to use this group of fields together more than for debugging.

scripts/taxonomy/SESARClassifierInput.py
```diff
@@ -0,0 +1,89 @@
+class SESARClassifierInput:
```
dannymandel, line 1: This looks great.  Thanks for making that change!

dannymandel (approved):

Issue #193, opened 2022-07-29
Add permit_id as a core model property


datadavev:
One or more permits are often associated with collection of physical samples, and there is a need to ensure that information is available with the sample records.

Task here is to add support for a multi-valued field to store permit ids with sample records. This should be a string field (i.e. not tokenized) with the assumption of storing some sort of identifier. In practice, it is likely values in this field will at least by quite variable since there's no formal representation of how this information is associated with records.

Initial use case is in support of GEOME and the FAIR Island project. Permit info will be added to the GEOME field `permitInformation`, an optional field that is attached to GEOME Event records. Some examples of entries in this field include:
```
1187/SU/KS/2006, 04239/ SU.3/KS/2019
LIPI 1187/SU/KS/2006 and 04239/SU.3/KS/2006
Commonwealth of Australia Torres Strait Fisheries Act 1984 Permit for Scientific purposes; Australian Fisheries Management Authority, Thursday Island
DAFF/DEA
NA
"Permit to EC and JD from Ministry of Fisheries and Forests, Export permit Permit C300/3019"
AU-COM2008042; Australian Government
```

We need to:
- Add a multi-valued string field to solr
- populate the field from GEOME records (should an attempt be made to parse what looks like multiple IDs?)
- Expose the field in isamples core records
- Support display of this field in the UI
- Support search against this field

datadavev:
Example spreadsheet with permit info provided by John in GDrive at: https://drive.google.com/drive/u/0/folders/1cIPzlGILRqG3Ip9477bsfJclCDpY1rVR

datadavev:
After further discussion with Neil, it seems there's a need for two properties: `complies_with`, and `authorized_by`.

datadavev:
Initial content will be added to GEOME records in the `permitInformation` field with the recommendation to use a structured format like:
```
permitInformation="complies_with:COMPLIES1;COMPLIES2 authorized_by:AUTHORIZED1;AUTHORIZED2"
```

Where:
* COMPLIESn is a pointer to the associated compliance documentation, ideally in the form of one or more URIs.
* AUTHORIZEDn is a pointer to the associated permit(s) authorizing the activity, ideally in the form of one or more URIs.

smrgeoinfo:
looks like the permitInformation is concatenated in the description field.  Is it important enough to add in a separate field on SamplingEvent or on Site?

datadavev:
[...]

PR #195, opened 2022-08-09
Geome permit information

dannymandel:
Initial work for https://github.com/isamplesorg/isamples_inabox/issues/193


isamples_metadata/GEOMETransformer.py
```diff
@@ -2,15 +2,20 @@
...
 import isamples_metadata
 from isamples_metadata.Transformer import (
     Transformer,
 )
 
+PERMIT_STRINGS_TO_IGNORE = ['nan', 'na', 'no data', 'unknown', 'none_required']
+
 TISSUE_ENTITY = "Tissue"
 JSON_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
+PERMIT_STRUCTURED_TEXT_COMPLIES_WITH_PATTERN = re.compile(r"complies_with:([^\s]+)(\s+authorized_by:(.*))?")
```
datadavev: Thinking about this a bit more- human entries are typically messy, especially during field work. Should probably also be prepared to accept tags like "compliesWith:", "complieswith:", maybe "complies with:", and perhaps even "complies:". Similarly for authorized_by.

tests/test_geome_permit_parsing.py
```diff
@@ -0,0 +1,88 @@
...
+        {"authorizedBy": [
+            "The SPICE project (Science for the protection of indonesian Coastal marine Ecosystems) was conducted and permitted under the governmental agreement between the German Federal Ministry of Education and Research (BMBF) and the Indonesian Ministry for Research and Technology (RISTEK), Indonesian Institute of Sciences (LIPI), Indonesian Ministry of Maritime Affairs and Fisheries (DKP), and Agency for the Assessment and Application of Technology (BPPT). This work was carried out in co-operation with Hassanuddin University (UNHAS, Makassar, Indonesia), Agricultural University Bogor (IPB, Bogor, Indonesia), and Jenderal Soedirman University (Purwokerto, Indonesia)."],
+            "compliesWith": []}),
+    ("DAFF/DEA", {"authorizedBy": ["DAFF", "DEA"], "compliesWith": []}),
+    ("Ministry of Environment, Natural Resources, and Transport",
```
datadavev, line 51: This one should result in a single string "`Ministry of Environment, Natural Resources, and Transport`"

datadavev (approved):
Good first pass. We will likely need to be a bit flexible on the acceptable text, but that might be something better done after exercising the current impl and getting feedback.

isamples_metadata/GEOMETransformer.py
dannymandel: All of these should be working now -- I've tweaked the regexes and added tests.

tests/test_geome_permit_parsing.py
dannymandel: I'm not sure how smart we can be here without regressing some of the other ones.  I'll remove the comma rule and see how many tests break and whether there is a way to tighten up that rule.

PR #197, opened 2022-08-11
First stab at conceptual documentation for solr relations

dannymandel:
@datadavev please let me know what else might be helpful here.  I had a difficult time writing the query section as I'm not sure what I wrote is actually how it is working.  Additionally, while revisiting the stuff I did last week, I'm not even sure It is doing what I thought it was doing last week.  Will fill in any gaps next week.  Thanks!


dannymandel:
@datadavev -- this is the doc I was mentioning this morning

Issue #198, opened 2022-08-16
Experimental: Implement a minimal Git backed, iSB harvestable record source.


datadavev:
There is a need for a simple system to support field data collection in an offline-first system that enables content creation and editing with no Internet connection, but is able to seamlessly participate in the iSamples network when a network connection becomes available.

This experiment will leverage CSV for file storage, git for revision control, GitHub for serving static content, and if necessary, supporting services operating within the constraints of zero-cost hosting such as on Deta, Vercel, Google App Engine, or AWS.

## Workflow

The workflow might be something like:

1. A GitHub repository is created based on a template that includes necessary actions and structure.
2. User creates a spreadsheet containing their data
3. User commits spreadsheet to local repo. Validation is run as part of the commit, or separately via script.
4. The spreadsheet is pushed to GitHub
5. A GH action parses the sheet and generates iSamples JSON entries for each record, an XML sitemap for the records, and publishes them on the `gh-pages` branch of the repository.
6. An iSB instance harvests and indexes the content.

## Some Questions

- Best place to assign identifiers? 
    - Ideally identifiers should be available at the time of content entry, so perhaps the repo includes a script that generates a spreadsheet template with requested number of pre-allocated identifiers?
- How far can we go with providing some sort of UI for the content, hosted on GH, zero-cost service, or perhaps just running locally with Docker etc?

dannymandel:
Presumably the identifiers would be allocated via datacite before they went to do fieldwork?

datadavev:
Yep, that would be best, though field work is typically a messy business. Another option may be to have the first part of the identifier available and unique to the sheet, and each row incrementing a suffix as needed. Then each record has a unique identifier but isn't registered until the sheet is committed. I'm not sure which is best from the user perspective.

dannymandel:
I think it will be less problematic overall to make registering N drafts a requirement before beginning the process.  

On the one hand, it makes more work up front but on the other it is going to save user headaches down the road by us not having to anticipate every manual algorithm someone could come up with.

datadavev:
Yep, that will definitely be less problematic for processing. We may need to adjust after some folks try the approach in the field though. Hard to say without testing where flexibility might be useful - so we can certainly go with the pre-registered approach first.

dannymandel:
[...]

PR #199, opened 2022-08-16
Initial data exploration/format

dannymandel:
Proposed CSV format for

https://github.com/isamplesorg/isamples_inabox/issues/198

Obviously not doing much yet, but verified the csv package works as expected here.  Wanted to get it out early for comments on format/structure etc.


tests/test_data/isb_core_csv_documents/simple_isamples.csv
```diff
@@ -0,0 +1,2 @@
...
+DOI:12345	Label 12345	This is a description of sample 12345.	["context1"]	["material1"]	["specimen1"]	["keyword1"]	["Homo sapiens"]	A label for the event that produced the sample.	A description of the event that produced the sample.	Something interesting about what produced the sample.	["Who is responsible for producing the sample"]	2022-01-01T00:00:00Z	A description of the sampling site.	The sampling site label.	2792	32.4434174	-110.7881222	["Geographical name 1", "Geographical name 2"]	["Sample registrant 1"]	["Sampling purpose 1"]	A special label for curation.	A description of how the sample is curated.	Any constraints associated with curation.	Where the sample is curated.	The institution responsible for curation.	["URI 1 to compliance documentation"]	["URI 1 to associated permit"]
```
dannymandel: Mostly just tab-separated strings -- exceptions are dates, numbers, and multi-valued strings, where we'll accept JSON arrays.  We could use a quote separator, but given that I knew we were going to embed multi-valued quoted string arrays as values, it seemed cleaner to use tabs.

dannymandel: The CSV module has a way to do custom dialects -- going to see if I can specify formats for the typed data and have it emit errors if not.

isb_lib/data_import/csv_import.py
```diff
@@ -59,14 +59,23 @@ def unflatten_csv_row(row: dict) -> dict:
         if k.startswith("producedBy"):
             flattened_row.pop(k)
             if "location" in k:
-                produced_by_sampling_site_location_dict[k.replace("producedBy_samplingSite_location_", "")] = v
+                k = k.replace("producedBy_samplingSite_location_", "")
+                if k == "elevationInMeters":
+                    k = "elevation"
```
dannymandel, line 64: @datadavev this looks like a difference between SOLR and our JSON format.  I prefer the more verbose `elevationInMeters` which we are using in SOLR, but I think we left it generic because we might want to use different units at some point.

isb_lib/data_import/csv_import.py
```diff
@@ -59,14 +59,23 @@ def unflatten_csv_row(row: dict) -> dict:
...
+                produced_by_sampling_site_location_dict[k] = v
             elif "samplingSite" in k:
                 produced_by_sampling_site_dict[k.replace("producedBy_samplingSite_", "")] = v
             else:
                 produced_by_dict[k.replace("producedBy_", "")] = v
         elif k.startswith("curation"):
             flattened_row.pop(k)
-            curation_dict[k.replace("curation_", "")] = v
+            if k == "curation_location":
+                k = "curationLocation"
```
dannymandel, line 73: This just seems overly verbose -- it ends up being 

```
curation: {
  "curationLocation": …
}
```

As opposed to 

```
  "curation": {
    "location": …
  }
```

isb_lib/data_import/csv_import.py
```diff
@@ -59,14 +59,23 @@ def unflatten_csv_row(row: dict) -> dict:
...
+            elif k == "curation_accessContraints":
+                k = "accessConstraints"
```
dannymandel, line 75: This is an unfortunate typo in the name of the solr field that I didn't notice until now…=\

datadavev: Happens...

isb_lib/data_import/csv_import.py
datadavev: Agreed. The latter makes sense to me.

isb_lib/data_import/csv_import.py
datadavev: Yes, elevationInMeters makes it quite clear. Let's use that.

isamples_metadata/GEOMETransformer.py
```diff
@@ -524,6 +524,10 @@ def parse_permit_freetext(text: str) -> dict[str, list[str]]:
 
         return GEOMETransformer._format_result_object([original_string])
```
dannymandel, line 526: This was a weirdness I found in the importer when turning on code coverage and running `pytest --cov` locally.  It was upset about `geo_to_h3` being in the module and was failing on the import.  The workaround was to move it inside the class called `GEOMETransformer` in the `GEOMETransformer` module.  This was some of the first python I wrote and didn't understand the distinction between classes and modules.  A long term fix would be to rename all the `XXXTransformer.py` to something like `xxx_transformer.py` in line with python recommended naming.

.github/workflows/python-app.yml
```diff
@@ -46,7 +46,7 @@
        working-directory: ./tests
        run: |
          source ../.venv/bin/activate       
-         pytest
+         pytest --cov --cov-fail-under=45
```
dannymandel: In my local testing, we had a total of ~46%, so as a first make sure we don't do *worse* when adding new code.  As we go along, we should be able to bump this up higher as well.

dannymandel:
I think we could continue mucking about with the format etc., but it seems like this is a good first step and ready to go.  We have multiple file formats working and tests for all of them.

PR #204, opened 2022-08-31
Try explicitly specifying pybind11 in pyproject.toml since inferred dependency resolution isn't working as expected

dannymandel:
Not sure what is happening here, I don't reproduce locally

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -41,6 +41,9 @@
        run: poetry install --no-interaction --no-root
      - name: Install
        run: |
+         poetry env info
+         poetry env use 3.9
+         poetry env info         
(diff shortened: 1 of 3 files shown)


dannymandel:
@datadavev Running on GitHub mysteriously broke from yesterday to today.  fasttext was the culprit for reasons I don't understand at all.  Ultimately the workaround involved manually running pip and installing fasttext and pybind11 before running poetry.  Additionally, disabling the poetry cache step helped.  This doesn't make much sense to me, but it unblocks our tests running on GitHub.  Once we get rid of fasttext I plan on undoing this change.

dannymandel:
@hyunssong FYI -- had nothing to do with your changes.  Please rebase your old PR on `develop` and open it again.  Thanks!

PR #214, opened 2022-09-13
New sitemaps test

dannymandel:
Was the last big piece of non-tested functionality to get the coverage up a bit more


dannymandel:
@datadavev FYI -- coverage up to 72% now.  This was the last piece of low-hanging fruit that was relatively easy to get incorporated into the test suite.  The good news is the bar will be much higher moving forward.  Ideally we'd be at ~80% for new code, but 72 is better than 64.

Issue #215, opened 2022-09-14
Integrate new metadata models


dannymandel:
We need to integrate the new metadata models that @hyunssong has been developing.  There will be multiple parts to this work.

dannymandel:
### iSB Repo
We need to incorporate the models into the iSB repo.  The current Smithsonian implementation should provide a general idea for how we want to incorporate the models.  We should create a new python module that will serve as a façade around each model.  The existing work lives in `isamplesfasttext.py`.  I suggest we make a new single module with multiple classes inside of it, one class per physical model file (or potentially one class per partner, managing multiple files if that is more appropriate).  We could call the file something like `metadata_models.py` underneath the `isamples_metadata` package.

Currently, the management of file paths is done directly inside the module by reading out of the `isamples_metadata.env` file.  It would be better to move all necessary config into the `isb_web_config.env` and the corresponding `config.py` in that package.  We'd then delete the files in `isamples_metadata`.  Additionally, we should change the implementation of the classes so that we may unit tests these new façade classes.  The best way to do this would be to push in the file path to each class that will wrap the model.  Then, we can provide stub implementations of the models and the corresponding paths to those stubs in a unit test for each model implementation.  Look at the example test data in `tests/test_data` for how to do this, e.g.

```
    local_url_path = f"file://{os.path.join(os.getcwd(), 'test_data/sitemaps')}"
    file_url = os.path.join(local_url_path, "test_sitemap_index.xml")
```

We'll need to instantiate the models initially when we start up the iSB server.  Look at the implementation of `on_startup()` in `main.py` for an example of how we connect to the database when we start.  It should be analogous to that.  We'll likely want to consolidate this logic into a method we can call from multiple places as well, since we'll need to call it the same way when we run the reindexing scripts:

```
@app.on_event("startup")
def on_startup():
    dao.connect_sqlmodel(isb_web.config.Settings().database_url)
```

so, there should be a one-line method call in there that will initialize all the models using the paths from config.

dannymandel:
### iSamples Docker Repo
The iSamples Docker repo will need to do two things with the models:
(1) Incorporate the config values into the container (I think this will be easy because they should mostly be static)
(2) Copy the model files into the container (should also be easy)

In looking at what we did for fasttext, there is a static version of the config file `isamples_metadata.env` that has this in it:

```
FASTTEXT_MODEL_PATH = "/app/sampledFeature.bin"
```

[...]

PR #216, opened 2022-09-17
Model integration

hyunssong:
based on #215. Mainly create a wrapper to load and call the classifier. I also had to move the scripting files (that were initially in scripts/taxonomy) due to import issues


isamples_metadata/SESARTransformer.py
```diff
@@ -283,7 +285,14 @@ def has_context_categories(self) -> typing.List[str]:
 
     def has_material_categories(self) -> typing.List[str]:
         material = self._material_type()
-        return MaterialCategoryMetaMapper.categories(material)
+        if not material:
+            # get the prediction result
+            categories = SESARMaterialPredictor.predict_material_type(
```
hyunssong: invoke sesar material model wrapper to predict when material field is empty

scripts/taxonomy/classification.py
```diff
@@ -11,15 +11,16 @@
...
     """Return the classification result"""
     # first pass : see if the record falls in the defined rules
     label = classify_by_rule(description_map, text, collection, labelType)
     if label:
         return (label, -1)  # set sentinel value as probability
     else:
-        # second pass : pass the record to the model
-        machine_prediction = classify_by_machine(text, collection, labelType)
+        # second pass : deriving the prediction by machine
+        # we pass the text to a pretrained model to get the prediction result
```
hyunssong: update the original script to take model as a parameter

isamples_metadata/metadata_models.py
```diff
@@ -1,6 +1,6 @@
 import os.path
 import logging
-from isamples_metadata import config
```
hyunssong: this is to combine the isampels_metadata's config into one config, isamples_web's config to make management more easy (see #215)

PR #217, opened 2022-09-21
Integrate models

hyunssong:
Based on #215. 
Update the current model classification scripts located in `scripts/taxonomy,` and call these scripts in the collections of `isamples_metadata`.


scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+            Prerequisite: In order to use this, make sure that there is a pydantic settings file on the
+            at the root of this repository named "isamples_web_config.env" with the config file path, and the model path
+            defined
+
+            :param collection : the collection type of the sample
+            :param label_type : the field that we want to predict
+        """
+        # load the model config file
+        if collection == "SESAR" and label_type == "material":
+            config_path = config.Settings().sesar_material_config_path
```
hyunssong: the config path here is the path of the  config.json that is used to load the metadata model . This is required to load the pretrained model, and I added these files to our google drive.

tests/.coveragerc
```diff
@@ -2,4 +2,5 @@
 omit=
     tests/*
     integration_tests/*
-		test*
\ No newline at end of file
+		test*
+    */scripts/*
```
hyunssong: remove scripts of classificaton from the test coverage. This is because we cannot test out these scripts as it is not possible to load the metadata models to git.

isb_web/config.py
```diff
@@ -69,6 +69,17 @@ class Settings(BaseSettings):
     # The authority id used for samples directly created in the iSB instance
     authority_id: str = "iSamples"
 
+    # The absolute path to the FastText model
+    fasttext_model_path: str = "UNSET"
```
hyunssong, line 73: we would want to integrate the .env file in the isamples_metadata into isb_web_config.env based on #215.

scripts/taxonomy/classification.py
```diff
@@ -68,6 +56,9 @@ def main(
     material_mapping = getHierarchyMapping("material")
     specimen_mapping = getHierarchyMapping("specimen")
 
+    MetadataModelLoader.initialize_models()
```
hyunssong: `classification.py` is a driver class that tests out the classifiers by applying it to database records.

scripts/taxonomy/Model.py
```diff
@@ -1,4 +1,3 @@
-import json
```
dannymandel: I recommend this module to a new `isamples_metadata/taxonomy` directory.  This looks to be part of the core model infrastructure as opposed to a script that exercises it.  The `scripts` folder should really just be a collection of cmd-line stuff, while the core functionality should be in one the main folders under `src`.

scripts/taxonomy/OpenContextClassifierInput.py
```diff
@@ -1,4 +1,4 @@
-from ClassifierInput import ClassifierInput
+from scripts.taxonomy.ClassifierInput import ClassifierInput
 
 
 class OpenContextClassifierInput(ClassifierInput):
```
dannymandel, line 4: I recommend this module move to a new `isamples_metadata/taxonomy` directory.  This looks to be part of the core model infrastructure as opposed to a script that exercises it.  The `scripts` folder should really just be a collection of cmd-line stuff, while the core functionality should be in one the main folders under `src`.

scripts/taxonomy/SESARClassifierInput.py
```diff
@@ -1,4 +1,4 @@
-from ClassifierInput import ClassifierInput
+from scripts.taxonomy.ClassifierInput import ClassifierInput
 
 
 class SESARClassifierInput(ClassifierInput):
```
dannymandel, line 4: I recommend this module move to a new `isamples_metadata/taxonomy` directory.  This looks to be part of the core model infrastructure as opposed to a script that exercises it.  The `scripts` folder should really just be a collection of cmd-line stuff, while the core functionality should be in one the main folders under `src`.

scripts/taxonomy/classification.py
dannymandel: (And this piece feels like it properly belongs in scripts)

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+class MetadataModelLoader:
+    """Class that instantiates the pretrained models"""
+
+    # loaded models that will be used for classification
+    _SESAR_MATERIAL_MODEL = None
+    _OPENCONTEXT_MATERIAL_MODEL = None
+    _OPENCONTEXT_SAMPLE_MODEL = None
+
+    @staticmethod
+    def load_model_from_path(collection, label_type):
```
dannymandel: Please add type information to the method parameters.

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+            logging.error(
+                "Unable to locate pretrained models",
+            )
+            return
+        with open(config_path) as json_file:
+            config_json = json.load(json_file)
+        # use the model config to get the pretrained model
+        model = Model(config_json)
+
+        # initialize the model fields
```
dannymandel: I would expect this to not re-instantiate the model and open the JSON every time it is called.  Is there somewhere else that prevents that from happening?

dannymandel: Ah, I see, it looks like `initialize_models` is responsible for that.

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+        if collection == "SESAR" and label_type == "material":
+            MetadataModelLoader._SESAR_MATERIAL_MODEL = model
+        elif collection == "OPENCONTEXT" and label_type == "material":
+            MetadataModelLoader._OPENCONTEXT_MATERIAL_MODEL = model
+        elif collection == "OPENCONTEXT" and label_type == "sample":
+            MetadataModelLoader._OPENCONTEXT_SAMPLE_MODEL = model
+
+    @staticmethod
+    def initialize_models():
```
dannymandel: I think the idea of this is good, but there's nothing that requires this method to be called.  The best solution might be to make each of the individual getter methods e.g. `get_xxx_model()` to do a lazy check to make sure it isn't `None` and then call `initialize_models` if it *is* `None`.  At a minimum we want some code that says you need to call `initialize_models()` before attempting to use this, or we'll silently fail.

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+    @staticmethod
+    def get_oc_sample_model():
+        return MetadataModelLoader._OPENCONTEXT_SAMPLE_MODEL
+
+
+class SESARMaterialPredictor:
+    """Material label predictor of SESAR collection"""
+    def __init__(self, model: Model):
+        self._model = model
+        self._model_valid = model is not None
```
dannymandel: Is there ever a valid use case for constructing a predictor with an invalid model?  If not, I recommend we fail with an exception on initialization.

dannymandel: It looks like `raise TypeError("Model is required to be non-None")` would do the trick.

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+        # get the prediction result with necessary fields provided
+        label, prob = self.classify_by_machine(input_string)
+        return (label, prob)
+
+
+class OpenContextSamplePredictor:
+    """Sample label predictor of OpenContext collection"""
+    def __init__(self, model: Model):
+        self._model = model
+        self._model_valid = model is not None
```
dannymandel: Same comment here on enforcing valid model.

tests/.coveragerc
dannymandel: Please revert this change.

dannymandel: Suggested test details described in https://github.com/isamplesorg/isamples_inabox/issues/215#issuecomment-1259806102

dannymandel: (And most of this code will move to a new package under `isamples_metadata`.

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+class SESARMaterialPredictor:
+    """Material label predictor of SESAR collection"""
+    def __init__(self, model: Model):
+        self._model = model
+        self._model_valid = model is not None
+        self._description_map = None
+
+    def checkInformative(self, text):
```
dannymandel: Please convert method name to `check_informative` per https://peps.python.org/pep-0008/#function-and-variable-names

I realize that not all of our code currently follows this but I've been trying to enforce it on newly added code.

Please add typing information to the text method parameter.

Once this is in `isamples_metadata`, I suspect `mypy` will start complaining about the lack of typing information.  I explicitly added type checking in https://github.com/isamplesorg/isamples_inabox/pull/218

scripts/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,318 @@
...
+                    informative = True
+                    break
+        if not informative and "sampleType" in self._description_map and \
+                text == self._description_map["sampleType"]:
+            informative = False
+        else:
+            informative = True
+        return informative
+
+    def checkInvalid(self, field_to_value):
```
dannymandel: Please rename to `check_invalid`

dannymandel (changes requested):
Overall this looks great @hyunssong!

I've added a few notes on suggested refactorings and the details on the test we need.  

Great work!

isamples_metadata/OpenContextTransformer.py
```diff
@@ -8,7 +8,7 @@
     StringEqualityCategoryMapper,
     AbstractCategoryMapper,
 )
-from scripts.taxonomy.metadata_models import (
+from isamples_metadata.taxonomy.metadata_models import (
```
hyunssong, line 11: moving files into `isamples_metadata`

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -55,55 +54,51 @@ def load_model_from_path(collection, label_type):
...
             MetadataModelLoader._OPENCONTEXT_SAMPLE_MODEL = model
 
-    @staticmethod
-    def initialize_models():
-        """
-            Invokes the load_model function to load all of the possible models
-            that are available based on the config
-        """
-        MetadataModelLoader.load_model_from_path("SESAR", "material")
-        MetadataModelLoader.load_model_from_path("OPENCONTEXT", "material")
```
hyunssong: This was defined as a method to be called from `isb_web/main.py` `on_startup()` to instantiate all the metadata models. But it seems that a better way would be instantiating the necessary models that we need through `load_model_from_path()` after checking if the model exists through `get_xxx_model()`. This update is based on the review, and the fact that not all isamples_boxes will have all the metadata models loaded. We expect that the metadata model that will be used within that box will only be loaded.

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -55,55 +54,51 @@ def load_model_from_path(collection, label_type):
...
-        MetadataModelLoader.load_model_from_path("OPENCONTEXT", "sample")
-
     @staticmethod
     def get_sesar_material_model():
+        if not MetadataModelLoader._SESAR_MATERIAL_MODEL:
```
hyunssong, line 82: checking whether the model exists and then initializes the model if not exists.

isb_web/main.py
```diff
@@ -100,8 +100,6 @@
 
 @app.on_event("startup")
 def on_startup():
-    # method to initialize all of the available models in config
-    MetadataModelLoader.initialize_models()
```
hyunssong: remove this and instead load the model by calling `get_xxx_model()` whenever we need to use it.

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -27,22 +28,25 @@ def load_model_from_path(collection, label_type):
 
             :param collection : the collection type of the sample
             :param label_type : the field that we want to predict
+            :param config_json : passed config json, if not passed we use the default path set in the .env
```
hyunssong, line 43: add the method to receive config_json for testing purposes

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        prediction, prob = self._model.predict(text)
+        return (
+            SESARClassifierInput.source_to_CV[prediction],
+            prob
+        )
+
+    def predict_material_type(
+        self, source_record: dict
+    ) -> Tuple[str, float]:
```
dannymandel: This is a bit of a nit, but it would be nicer if this returned a simple class like `PredictionResult` that looked something like:

```
class PredictionResult:
    def __init__(self, value: str, confidence: float):
        self.value = value
        self.confidence = confidence
```

this way, callers won't need to go look at the code that's returning it to figure which index in the tuple is the value and which is the confidence.  Just makes it a little easier for people to work with.

dannymandel: It looks like all the of different prediction methods return roughly the same thing, so we could use the same class everywhere.

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+class OpenContextMaterialPredictor:
+    """Material label predictor of OpenContext collection"""
+    def __init__(self, model: Model):
+        if not model:
+            raise TypeError("Model is required to be non-None")
+        self._model = model
+
+    def classify_by_machine(self, text: str):
```
dannymandel: Missing a return type hint on this method

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,305 @@
...
+        # check if the fields fall into the rules
+        # if it does not, return None
+        result = self.classify_by_sample_type(field_to_value)
+        if result:
+            # map to controlled vocabulary
+            return SESARClassifierInput.source_to_CV[result]
+        else:
+            return None
+
+    def classify_by_machine(self, text: str):
```
dannymandel: Missing a return type hint on this method

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -0,0 +1,305 @@
...
+        if field_to_value["sampleType"] == "U-channel":
+            return "Sediment"
+        if field_to_value["sampleType"] == "CTP":
+            return "Liquid"
+        if field_to_value["sampleType"] == "Individual Sample>Cylinder":
+            return "Material"
+        # if the record cannot be classified by the defined rules
+        return None
+
+    def classify_by_rule(self, text: str, description_map: dict):
```
dannymandel: Missing a return type hint on this method

tests/test_metadata_models.py
```diff
@@ -0,0 +1,133 @@
+from isamples_metadata.taxonomy.metadata_models import (
```
dannymandel, line 1: This test looks great.  Thanks for writing it!

scripts/taxonomy/classification.py
```diff
@@ -74,12 +74,12 @@ def main(
...
             # load the model predictor
             smp = SESARMaterialPredictor(sesar_model)
-            label, prob = smp.predict_material_type(
+            result = smp.predict_material_type(
                 thing.resolved_content
             )
 
             print(
-                f"Predicted (probability, label) : {label}, {prob}"
+                f"Predicted (probability, label) : {result.value}, {result.confidence}"
```
dannymandel, line 82: This is really nit, but for future reference you can override `__repr__` on the class itself like so:

```
 def __repr__(self):
        return f"{result,value}, {result.confidence}"
```

then the callers would just need

`f"Predicted (probability, label) : {result}"`

dannymandel (approved):
Looks great!  Thanks for taking the time to implement those changes @hyunssong

PR #218, opened 2022-09-27
Make metadata a typed package

dannymandel:
This should have been here before -- we should be running `mypy` on this.


dannymandel:
Simple change to make sure that `isamples_metadata` has proper type annotations.

dannymandel:
@datadavev @hyunssong FYI -- something I encountered while doing the review for https://github.com/isamplesorg/isamples_inabox/pull/217/files

dannymandel:
(And we do want to make sure all new code has typing annotations)

PR #220, opened 2022-10-12
Update frictionless submodule to one that has the necessary extras

dannymandel:
Merge back in changes we made yesterday to get the app going


dannymandel:
@datadavev FYI

PR #223, opened 2022-10-14
Make sure we check out submodules when setting up our integration test environment

dannymandel:
After all that back and for this morning, @datadavev, I think this was the *actual* cause of the integration test failing.  Maybe there were multiple, I'm not sure, but I merged that other PR on Wednesday and I think that made things fail.


dannymandel:
The test failure makes no sense to me -- it is failing on something that passes locally, and hasn't failed any other time.

PR #225, opened 2022-10-21
Sesar indexing fixes

hyunssong:
Updates to make SESAR indexing to work. Created a new set of test records (as the previous one had wrong igsns), made sure that when extracting text fields from the record is done in a consistent manner, update codes to allow Top3 high confidence results are going to be returned.


isamples_metadata/taxonomy/OpenContextClassifierInput.py
```diff
@@ -11,14 +11,15 @@ def __init__(self, thing):
         self.OC_material_field = "Consists of_label"
         self.OC_sample_field = "Has type_label"
 
-    def build_text(self, description_map, labelType):
+    def build_text(self, description_map, labelType, description_field):
         """Return the concatenated text of informative fields in the
         description_map"""
         concatenated_text = ""
-        for key, value in description_map.items():
+        for field in description_field:
```
hyunssong, line 18: this is going to be further updated in a separate branch where I will add updates for OpenContext.

isamples_metadata/taxonomy/SESARClassifierInput.py
```diff
@@ -79,20 +80,24 @@ def __init__(self, thing):
...
-                # empty field
-                continue
-            elif labelType == "material" and key != self.SESAR_material_field:
-                concatenated_text += value + " , "
-            elif labelType == "sample" and key != self.SESAR_sample_field:
-                concatenated_text += value + " , "
+        # use the order of the fields to build text
+        # to have consistent text content
+        for field in field_order:
+            if field in description_map:
```
hyunssong, line 89: Instead of extracting text fields from the dictionary `description_map,` fix the order of the fields that are going to be extracted using `field_order` when building the text.

isamples_metadata/taxonomy/Model.py
```diff
@@ -29,11 +29,14 @@ def predict(self, text):
...
-        predicted_class = predicted_class.cpu().item()
-        prob = logits.softmax(dim=-1)[0][predicted_class].item()
-        return (
-            self.config["CLASS_NAMES"][predicted_class],
-            prob
-        )
+        # convert logits into probability
+        prob  =  logits.softmax(dim=-1)[0]
+        # get the top 3 high confidence predictions
+        top_probs, indices = torch.topk(prob, 3)
```
hyunssong, line 35: Display the top3 high confidence labels using `topk`. This extracts the highest 3 probable labels.

isamples_metadata/taxonomy/metadata_models.py
```diff
@@ -211,11 +211,11 @@ def classify_by_machine(self, text: str) -> Tuple[str, float]:
...
         """
-        prediction, prob = self._model.predict(text)
-        return (
-            SESARClassifierInput.source_to_CV[prediction],
+        predictions = self._model.predict(text)
+        return [(
+            SESARClassifierInput.source_to_CV[label],
             prob
-        )
+        ) for (label, prob) in predictions]
```
hyunssong, line 218: updating to list structure as multiple tuples will be returned.

scripts/taxonomy/sesar_indexing_test_records.json
```diff
@@ -0,0 +1,3166 @@
+{
```
dannymandel: Could you delete this file and replace the lines in `isamples_inabox/integration_tests/test_model_values.json` please?

isamples_metadata/SESARTransformer.py
```diff
@@ -309,7 +309,7 @@ def has_material_categories(self) -> typing.List[str]:
             sesar_model = MetadataModelLoader.get_sesar_material_model()
             # load the model predictor
             smp = SESARMaterialPredictor(sesar_model)
-            categories = [smp.predict_material_type(self.source_record).value]
+            categories = [prediction.value for prediction in smp.predict_material_type(self.source_record)]
```
dannymandel, line 312: Do we expect these to actually return multiple useful values for SESAR?  Or do we still expect one?

hyunssong: I expect this will return multiple labels (top 3 prediction labels).

hyunssong: Thinking again, I think there might be an issue here. The test record currently contain the list of high confidence labels, and the confidence of each of such prediction. But the `has_material_category` only returns a list of labels. I think we might need an additional field to store confidence scores in future, if we want to compare those confidence values.

dannymandel: Yeah, for now we are ignoring the confidence values in the test because we don't have them stored anywhere.

dannymandel: (And confirmed that the tests pass with the multiple values)

dannymandel (approved):
Looks good and I verified that the tests are passing now, as well.  Thanks @hyunssong!

PR #228, opened 2022-10-26
Confidence fields for indexing

dannymandel:


integration_tests/test_solr_search_results.py
```diff
@@ -169,7 +171,8 @@ def _transformed_json_to_test_tuples() -> list[tuple]:
             current_tuple = (
                 key,
                 {
-                    "hasMaterialCategory": value.get("material").get("values")
+                    "hasMaterialCategory": value.get("material").get("values"),
+                    "hasMaterialCategoryConfidence": value.get("material").get("confidence")
```
dannymandel, line 175: We now check both the indexed confidence values and the indexed string values

isamples_metadata/SESARTransformer.py
```diff
@@ -301,17 +305,39 @@ def has_context_categories(self) -> typing.List[str]:
             material_type, primary_location_type
         )
 
-    def has_material_categories(self) -> typing.List[str]:
-        material = self._material_type()
-        if not material:
+    def _compute_material_prediction_results(self) -> typing.Optional[typing.List[PredictionResult]]:
```
dannymandel, line 308: Need to refactor this into something that will work generally across multiple category/confidence pairs.  Will do that work once we have the next model combo to integrate.

isamples_metadata/Transformer.py
```diff
@@ -24,6 +24,10 @@ class Transformer(ABC):
 
     N2T_ARK_NO_HTTPS_PREFIX = f"{N2T_NO_HTTPS_PREFIX}ark:/"
 
+    RULE_BASED_CONFIDENCE = 1.0
```
dannymandel, line 27: New confidence values as discussed on Wednesday -- 0…1 is machine-based, 1.0 is rule-based, 2.0 is human-entered.

isamples_metadata/Transformer.py
```diff
@@ -134,21 +144,42 @@ def sample_sampling_purpose(self) -> str:
...
+            return None
+        confidences = []
+        for _ in category_list:
+            confidences.append(Transformer.RULE_BASED_CONFIDENCE)
+        return confidences
+
     @abstractmethod
     def has_context_categories(self) -> typing.List[str]:
         """Map from the source record into an iSamples context category"""
         pass
```
dannymandel, line 160: New default implementations for all confidences -- will set things to 1.0 unless overridden.

scripts/migrations/solr/add_confidence_values.py
```diff
@@ -0,0 +1,74 @@
+import logging
```
dannymandel, line 1: Solr migration to populate existing records with confidence values.

solr_schema_init/create_isb_core_schema.py
```diff
@@ -128,8 +128,11 @@ def createCollection():
 createCopyField("description", "description_text")
 createCopyField("description", "searchText")
 createField("hasContextCategory", "string", True, True, None, True)
```
dannymandel, line 130: New solr fields to store the confidences.

datadavev (approved):
This all looks good. In general I think it would be good to have less code dependency on model property values (like the vocab terms), but the current approach works fine for now.

dannymandel:
Could you explain a bit more about what you mean by that?  I'm not sure I understand.

datadavev:
So for example MaterialCategoryMetaMapper hard codes different types of mappers tied to specific values. Could instead load the mappers load from a config like a json file, basically shifting the property specific values to config rather than code. The config might have entries like:
```json
[
  {
    "type":"endsWith",
    "input":["Rock"],
    "result":"Rock"
  },
  {
    "type":"equals",
    "input":["Siderite>Mineral", "Macrobiology>Other", "Organic Material"],
    "result":"Organic Material"
  }
]
``` 

Then MaterialCategoryMetaMapper might have something like loadJson() that populates its list of mappers. 

It shifts a bunch of stuff to config rather than code, but still functions the same so really the current approach is fine.

dannymandel:
Got it.  Yeah, that makes sense and would be nice.  It might lower the barrier to entry for other people to write transformers.

PR #229, opened 2022-11-01
Integrate noidy into iSB

dannymandel:
Get the bots running here, not ready for primetime


datadavev (commented):
This all looks good so far.

dannymandel:
@datadavev -- I think this one is more or less complete now, and should be up to parity with our previous DataCite implementation.  

An obvious next step might be to integrate it into the GUI app for offline use -- the question then would be how would we reconcile the minter state between the iSB database and the offline copy that might be running out in the field.

Anyhow, I think it's at a good stopping point now, and would appreciate guidance on next steps.

isb_lib/models/namespace.py
```diff
@@ -0,0 +1,73 @@
...
+            doc="Internal state of the minter associated with this namespace",
+        ),
+    )
+
+    def _allowed_people_copy(self) -> list[str]:
+        if self.allowed_people is not None:
+            return self.allowed_people.copy()
+        else:
+            return []
+
```
dannymandel, line 60: I suspect there is some way to make our `StringListType` properly mark itself dirty in order to avoid this silliness, but I wasn't able to locate it in a quick perusal of the docs.  This approach isn't terrible, just inelegant.

isb_web/manage.py
```diff
@@ -154,12 +153,12 @@ async def __call__(
 )
 
 
-class MintIdentifierParams(BaseModel):
```
dannymandel, line 157: Rename all the previous implementations to reflect the fact that they're DataCite specific.

tests/test_noidy.py
```diff
@@ -0,0 +1,21 @@
...
+    _mint_identifiers_and_assert(minter, num_identifiers)
+
+
+def _mint_identifiers_and_assert(minter, num_identifiers):
+    identifiers = minter.mint(num_identifiers)
+    returned_identifiers = 0
+    for identifier in identifiers:
+        returned_identifiers += 1
+        assert len(identifier) > 0
+    assert returned_identifiers == num_identifiers
```
datadavev: Should probably add in a check for uniqueness, e.g:
```python
idset = set(identifiers)
assert len(idset) == len(identifiers)
```

datadavev (approved):

PR #231, opened 2022-11-08
Dynamic import test

dannymandel:
See if this works when run in CI…


datadavev (approved):
I like the simplicity of this solution

PR #236, opened 2022-11-18
Compute h3 hashes at different resolutions

### LINKED ISSUE #235: Add solr fields for different H3 resolutions
It would be beneficial to rapidly compute the spatial distribution of records at different resolutions. This could be used for example, to augment the spatial display in the UI which is currently very slow when viewing broad regions (since it is dependent on streaming individual records) and limited to a reasonable number (e.g. 50k) records in a view.

A spatial distribution / heatmap can be very efficiently computed by faceting on the h3 values, however this requires that the h3 strings are available at specific resolutions (i.e. length of the h3 string) since solr is not able to compute facets on the first n chars of a string. We currently store only full resolution h3 in `producedBy_samplingSite_location_h3`, e.g. `8f893424431b471`

Task here is to add fields for the h3 resolutions 1 through 10, which will enable rapid distribution display at resolutions from global down to a few km2. 

The requirements for these fields are: facet on field, search on field. The [Solr docs](https://solr.apache.org/guide/solr/latest/indexing-guide/field-properties-by-use-case.html) suggest`docValues=true`, `indexed=false`, `stored=false`, `multiValued=false`. e.g.:

```
<field name="producedBy_samplingSite_location_h3_1" type="string" indexed="false" stored="false" docvalues="true />
```

Fields can be named like the full resolution, but with a suffix `_n` where `n` is `1`, `2`, ... `10`. The resulting field `producedBy_samplingSite_location_h3_1` would have two characters, e.g. `8f` from the example above.

These fields could be populated as copyFields, e.g.:
```
<copyField source="producedBy_samplingSite_location_h3" dest="producedBy_samplingSite_location_h3_1" maxChars=2" />
<copyField source="producedBy_samplingSite_location_h3" dest="producedBy_samplingSite_location_h3_2" maxChars=3" />

```

dannymandel:
Fixes https://github.com/isamplesorg/isamples_inabox/issues/235


isamples_metadata/Transformer.py
```diff
@@ -113,6 +115,10 @@ def transform(self) -> typing.Dict:
             "authorizedBy": self.authorized_by(),
             "compliesWith": self.complies_with(),
         }
+        for index in range(1, 11):
```
datadavev: I think we may need to add one more level at resolution 0, so `range(0,11)`

datadavev (approved):
One minor adjustment to add resolution 0 but otherwise looks good. Initial testing shows excellent performance

PR #239, opened 2022-12-07
API endpoint for h3 match counting 

### LINKED ISSUE #237: API endpoint for h3 match counting
We need an API that can return the number of documents given one or more H3 values.

Something like:

```python
counts_by_h3_indices(h3s:set[str], q:str="*:*") -> dict
```
The returned dict has keys of the provided h3 values, and values being the number of documents that match the query `q` and the h3 value.

Note that it will be necessary to compute the appropriate h3 field by getting the resolution of each h3 value (https://uber.github.io/h3-py/api_reference.html#h3.get_resolution).

Not sure if there's a way to do this in solr without simply issuing a query per h3 value.

dannymandel:
Initial fixes for https://github.com/isamplesorg/isamples_inabox/issues/237


dannymandel:
Fixes https://github.com/isamplesorg/isamples_inabox/issues/237

dannymandel:
New API available at `/h3_counts/`

PR #244, opened 2023-01-03
Disable auto field creation, as this may create the wrong field type and leave the index in an unusable state

dannymandel:
Make it impossible to use the harmful `autoCreateFields` mode.  In local testing I observed (after initial creation), when hitting http://localhost:8984/solr/isb_core_records_3/config:

```
"updateRequestProcessorChain":[{
        "default":"true",
        "name":"add-unknown-fields-to-the-schema",
        "processor":"uuid,remove-blank,field-name-mutating,parse-boolean,parse-long,parse-double,parse-date,add-schema-fields",
        "":[{"class":"solr.LogUpdateProcessorFactory"},
          {"class":"solr.DistributedUpdateProcessorFactory"},
          {"class":"solr.RunUpdateProcessorFactory"}]}],
```

Then after running the new code, I observed

```
"updateRequestProcessorChain":[{
        "default":"false",
        "name":"add-unknown-fields-to-the-schema",
        "processor":"uuid,remove-blank,field-name-mutating,parse-boolean,parse-long,parse-double,parse-date,add-schema-fields",
        "":[{"class":"solr.LogUpdateProcessorFactory"},
          {"class":"solr.DistributedUpdateProcessorFactory"},
          {"class":"solr.RunUpdateProcessorFactory"}]}],
```

So the "add-unknown-fields-to-the-schema" has flipped from default `true` to default `false`.

### DIFF
--- a/solr_schema_init/create_isb_core_schema.py
+++ b/solr_schema_init/create_isb_core_schema.py
@@ -6,8 +6,11 @@ import time
 # Sleep for 10 seconds to allow for the Solr Docker container to start up.
 time.sleep(10)
 
-CREATE_COLLECTION_API = "http://solr:8983/solr/admin/collections?action=CREATE&name=isb_core_records&numShards=1&replicationFactor=1"
-SOLR_API = "http://solr:8983/api/collections/isb_core_records/"
+COLLECTION_NAME = "isb_core_records"
+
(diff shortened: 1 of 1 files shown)


datadavev (commented):
Looks good to me

PR #246, opened 2023-01-04
Fetch raw columns for solr import

dannymandel:
Incomplete experiment to see if only fetching individual columns could speed up the solr import slowness we're seeing.  Results on my local machine indicate there are minimal gains to be had, if any.


dannymandel:
@datadavev FYI

dannymandel:
Closing this PR as it isn't effective.

Issue #247, opened 2023-01-10
Support for "reliquery" documents


datadavev:
The notion of a "reliquery" document is to provide a mechanism for a user to conveniently cite a potentially large number of samples so that others may unambiguously discern the set of records being cited.

There is currently no defined form for such a document, though it will likely contain:

- timestamp 
- the system on which the subset is being gathered from (server url)
- the query used to identify the subset of records
- the number of records matching the query
- the number of record identifiers returned in the reliquery document
- a list of identifiers matching the query
- an optional user provided description of the document

This should be a JSON document.

This document should be produced through an API request that accepts as input the query and the optional description.

For prototyping, a maximum of 10,000 identifiers should be sufficient. 

Anticipated future enhancements would be to include a list of appropriate citations for the samples grouped by source authority. For example, all Smithsonian records may fall under a single citation. Other sources may require grouping at different levels such as author or creator of the samples. This aspect is undefined at this stage but listed here since it implies potential for possibly significant additional processing to be applied when generating the reliquery document.

dannymandel:
```
{
  "timestamp": "2023-01-10T15:05:34.721295Z",
  "url": "http://localhost:8000/thing/reliquery",
  "query": "source:SESAR",
  "count": 4606504,
  "return_count": 10000,
  "identifiers": [
    "IGSN:SLB0000GZ",
    "IGSN:SLB0000H0",
    "IGSN:JBG00006B",
    …
  ],
  "description": "This is a test description of the test query"
}
```

PR #248, opened 2023-01-10
Initial reliquery support

dannymandel:
Initial implementation of https://github.com/isamplesorg/isamples_inabox/issues/247

### DIFF
--- a/isb_web/api_types.py
+++ b/isb_web/api_types.py
@@ -1,0 +1,28 @@
+from typing import Optional
+
+from pydantic import BaseModel
+
+
+class ThingsSitemapParams(BaseModel):
+    identifiers: list[str]
+
+
+class ReliqueryResponse(BaseModel):
+    timestamp: str
+    url: str
+    query: str
+    count: int
+    return_count: int
+    identifiers: list[str] = []
+    description: Optional[str]
+    # The following fields are only present if a query is re-executed:
+    previous_timestamp: Optional[str]
+    previous_url: Optional[str]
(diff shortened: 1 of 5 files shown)


isb_web/api_types.py
```diff
@@ -0,0 +1,59 @@
+from typing import Optional
+
+from pydantic import BaseModel
+
+from isb_web.isb_enums import ISBAuthority, ISBReturnField
+
+"""
+Module for various API parameter and return types used throughout iSamples API
+"""
```
dannymandel, line 9: Moved this out to its own file as `main.py` has grown a bit long in the tooth and this seemed like a clean way of organizing things while trimming down a bit of that cruft.

isb_web/main.py
```diff
@@ -269,6 +270,25 @@ async def get_solr_select(request: fastapi.Request):
...
+        identifiers.append(dict["id"])
+    reliquery_response = ReliqueryResponse(
+        query=params.query,
+        identifiers=identifiers,
+        timestamp=datetime.datetime.now().strftime(SOLR_TIME_FORMAT),
+        count=json_response.get("numFound"),
+        return_count=len(identifiers),
+        url=str(request.url),
+        description=params.description)
+    return reliquery_response
```
dannymandel, line 300: I tried to make it so you could just include the old response in the payload for this method, but that seemed to break if the old response wasn't present.  I'll need to mess around with it a bit more tomorrow.

datadavev (approved):

PR #252, opened 2023-01-25
Switch to DateDataParser for performance reasons

dannymandel:
Upon examining the indexing performance, it seemed like there were many times when we were in the dateparser stacktrace.  Dateparser is known to be a bit slow, so I looked at their documentation, and they suggested DateDataParser as an alternative: https://dateparser.readthedocs.io/en/latest/usage.html

I wrote a simple test on my Mac, and for 5000 dates, `dateparser.parse` was ~2200ms and `DateDataParser` was ~1600ms.  Not super fast either way, but if we extrapolate that out to 6000000 calls, it could easily save a few minutes.  We had unit tests in place for the date calls, and they all pass unmodified with the new api calls in place.

### DIFF
--- a/isb_lib/core.py
+++ b/isb_lib/core.py
@@ -14,7 +14,7 @@ import igsn_lib.time
 from isamples_metadata.metadata_exceptions import MetadataException
 from isb_lib.models.thing import Thing
 from isamples_metadata.Transformer import Transformer
-import dateparser
+from dateparser.date import DateDataParser
 import re
 import requests
 import shapely.wkt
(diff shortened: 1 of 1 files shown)


datadavev:
certainly a worthwhile change.

We could also spend some time to identify the specific patterns we expect to deal with and use the built in date parser and perhaps fall back to datadateparser if necessary

dannymandel:
I think that would be a useful thing to do as well, but I wasn't as certain that I wouldn't regress functionality with that approach.  The nice thing about dateparser is that it is very flexible, and we know our input is (putting it politely) flexible as well.

dannymandel:
I guess I could make the regexes very tight, and it wouldn't break anything, only speed it up.  I'll spend a couple minutes and see what I can get…

dannymandel:
I did this

```
DATE_PATTERN = re.compile(r"^(\d\d\d\d)(?:-(\d\d))?(?:-(\d\d))?$")

    date_match = DATE_PATTERN.fullmatch(raw_date_str)
    if date_match is not None:
        year = date_match.group(1)
        if date_match.group(2) is not None:
            month = date_match.group(2)
        else:
            month = 1
        if date_match.group(3) is not None:
            day = date_match.group(3)
        else:
            day = 1
        date_time = datetime.datetime(year=int(year), month=int(month), day=int(day), tzinfo=UTC)
    else:
        date_time = ddp.get_date_data(raw_date_str, date_formats=RECOGNIZED_DATE_FORMATS)
```
That matched things like `1985`, `1985-07`, and `1981-01-01`.  But it really didn't speed much up. I don't think it's worth taking it.

PR #257, opened 2023-01-27
Initial implementation of GBIF taxon name cache

dannymandel:
This is the code that builds the cache and db schema for storing it -- to be implemented: the code that uses this cache on a GEOME import.  It remains to be seen how well this will work.  I did note that their API does fuzzy matching, so we'll need to see how well this works in practice, but it should be straightforward to check.


isb_lib/models/taxonomy_name.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
+
+import sqlalchemy
+from sqlmodel import SQLModel, Field
+
+
+class TaxonomyName(SQLModel, table=True):
+    primary_key: Optional[int] = Field(
+        # Need to use SQLAlchemy here because we can't have the Python attribute named _id or SQLModel won't see it
+        sa_column=sqlalchemy.Column(
```
datadavev, line 10: does gbif provide a unique value / id for each name? if so then there may be some value in using that here. Just thinking through some potential other uses of the cache.

datadavev: nm. I see the db is populated as a simple list of names, using both sci name and possibly canonical name, hence a gbif id wouldn't help

datadavev (approved):

isb_lib/models/taxonomy_name.py
dannymandel: We certainly could add the id as a column and use the same id for both sci name and canonical.

dannymandel:
Had time to circle back on this one today and wrote the solr migration script to see how much of the data we could cover with it.  Ran it over the geome things in the db and got this output:

```
2023-03-28T16:59:11 root:INFO: Visited 271000 things, current percentage of things with identified kingdom: 0.9653210332103321
2023-03-28T16:59:11 root:INFO: missing values for {'Animal', 'NA', 'unknown', 'unrecorded', 'crinia', 'Metazoa', 'Dikaryomycota', 'Unknown', 'litoria', 'blank', 'GET', 'not applicable'}
```

In short, that's pretty good coverage!

dannymandel:
Going to merge this and get it deployed on iSC AWS so we can build the cache there.

PR #258, opened 2023-02-01
Point the integration tests at the AWS instance

dannymandel:
Have these run against the deployed AWS instance rather than Hyde


dannymandel:
@datadavev -- FYI I ran these locally against the AWS instances and they worked so I'm pointing them there instead of at Hyde

PR #260, opened 2023-02-15
Solr load balancing

dannymandel:
A very simple implementation of a solr load balancer that picks from a random solr live instance.

TODO: refresh the list over time (in case one went down, for instance)


dannymandel:
Don't recommend taking this.  It didn't end up helping performance.

PR #261, opened 2023-02-21
From the python side, this is all that is required to use gunicorn

dannymandel:
Hi @datadavev this is what would be required on the ISB side if we wanted to turn on gunicorn.  Note that it didn't actually help performance in the experiments.  Also, some of the uvicorn static file rewrites didn't seem to function properly under gunicorn (I didn't investigate as the API calls were working, and that is what I was testing)


dannymandel:
See also https://github.com/isamplesorg/isamples_docker/pull/38

dannymandel:
Don't recommend taking this.

dannymandel:
Closing this one.

PR #263, opened 2023-02-21
experimental change to see if reusing session helps matters

dannymandel:
FWIW, this seemed to make things slower


dannymandel:
Don't recommend taking this change.

PR #265, opened 2023-03-01
Migration scripts to generate all the model-based output for OpenContext and SESAR vocabulary

dannymandel:
A couple scripts to generate `.tsv` files that will be processed by a later script that will update the solr index.


dannymandel:
Going to merge so I can deploy this out to AWS and start the scripts running

PR #267, opened 2023-03-22
New migration script to ingest the pre-computed values obtained from running the model

dannymandel:
This has been run against the AWS solr


dannymandel:
@datadavev FYI -- this is the script we used to load up the model output into Solr

PR #269, opened 2023-03-29
Use taxon cache in transform

dannymandel:
Follow up to https://github.com/isamplesorg/isamples_inabox/pull/257

That populated the database cache (and wrote a solr migration), this integrates it into every transform request


dannymandel:
@datadavev FYI, this is integrated now

PR #271, opened 2023-05-10
Fixes for problems observed when we pushed prometheus out

dannymandel:
Assorted weirdness fixes

### DIFF
--- a/isb_web/main.py
+++ b/isb_web/main.py
@@ -67,7 +67,6 @@ app = fastapi.FastAPI(openapi_tags=tags_metadata)
 dao = SQLModelDAO(None)
 manage_app = manage.manage_api
 debug_app = debug.debug_api
-metrics_app = metrics.metrics_api
 # Avoid a circular dependency but share the db connection by pushing into the various handlers
 manage.dao = dao
 metrics.dao = dao
@@ -102,7 +101,7 @@ app.mount(
 )
 app.mount(manage.MANAGE_PREFIX, manage_app)
 app.mount(debug.DEBUG_PREFIX, debug_app)
-app.mount(metrics.METRICS_PREFIX, metrics_app)
+app.include_router(metrics.router)
 
 
 @app.on_event("startup")
--- a/isb_web/metrics.py
+++ b/isb_web/metrics.py
@@ -2,15 +2,14 @@ import logging
 import time
 from typing import Optional
 
-from fastapi import FastAPI
+from fastapi import APIRouter
 from starlette.responses import PlainTextResponse
 
 from isb_web.isb_solr_query import solr_counts_by_authority
 from isb_web.sqlmodel_database import SQLModelDAO, things_by_authority_count_dict
 
-metrics_api = FastAPI()
+router = APIRouter(prefix="/metrics")
 dao: Optional[SQLModelDAO] = None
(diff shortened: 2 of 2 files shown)


isb_web/main.py
```diff
@@ -102,7 +101,7 @@
 )
 app.mount(manage.MANAGE_PREFIX, manage_app)
 app.mount(debug.DEBUG_PREFIX, debug_app)
-app.mount(metrics.METRICS_PREFIX, metrics_app)
+app.include_router(metrics.router)
```
dannymandel, line 104: Using this seems to work around the `root_path` problems where things were redirecting to the wrong place.

dannymandel: Note that this seems like a cleaner approach overall that doing the embedded app approach.  We should probably convert the `manage_app` and `debug_app` implementations at some point.

isb_web/metrics.py
```diff
@@ -2,15 +2,14 @@
...
 
-from fastapi import FastAPI
+from fastapi import APIRouter
 from starlette.responses import PlainTextResponse
 
 from isb_web.isb_solr_query import solr_counts_by_authority
 from isb_web.sqlmodel_database import SQLModelDAO, things_by_authority_count_dict
 
-metrics_api = FastAPI()
+router = APIRouter(prefix="/metrics")
```
dannymandel, line 11: This is the preferred way (according to FastAPI docs) to mount subpaths underneath an app.

isb_web/metrics.py
```diff
@@ -56,3 +54,15 @@ def root():
     metrics.solr_counts = solr_counts_by_authority()
     metrics.solr_scrape_duration_seconds = time.time() - db_end_time
     return PlainTextResponse(metrics.metrics_string())
+
```
dannymandel, line 57: This part is a little goofy -- I couldn't figure out why things were behaving different for `/metrics` vs. `/metrics/` requests.  In the end I threw my hands up and just explicitly handled both of them.  I did look in the docs if there was a way to just configure this so I didn't have to declare separate paths, but I didn't find anything.

isb_web/main.py
datadavev: It seems that `include_router` is the way to work with an application composed of multiple files[^1] whereas `mount` is for combining independent applications[^2]. 

I think the `include_router` approach makes sense here since these are really parts of the same application, just on different routes, rather than separate applications.

[^1]: https://fastapi.tiangolo.com/tutorial/bigger-applications/
[^2]: https://fastapi.tiangolo.com/advanced/sub-applications/

isb_web/metrics.py
datadavev: Yep, this looks good

datadavev (approved):
Changes make sense and we should adjust the app structure as appropriate. Using the router approach might simplify things like db connection management.

isb_web/metrics.py
datadavev: could be that `/metrics` is getting redirected to `/metrics/`?

dannymandel: Yes, that is definitely what was happening, and for whatever reason prometheus was choking on that.  It seemed like an undesirable behavior, so I just made both work with no redirect.

PR #273, opened 2023-05-19
add auth code for annotation post

hyunssong:
Adding authentication code to handle creating annotations using throughput.

### DIFF
--- a/isb_web/manage.py
+++ b/isb_web/manage.py
@@ -204,7 +204,11 @@ async def login(request: starlette.requests.Request):
     Initiate OAuth2 login with ORCID
     """
     redirect_uri = request.url_for("auth")
-    return await oauth.orcid.authorize_redirect(request, redirect_uri)
+    # check if login is for annotation purpose, if so add query param 
+    if "annotation" in request.query_params and request.query_params["annotation"]=="true" :
+        redirect_uri += "?annotation=true"
+    orcid_auth_uri = await oauth.orcid.authorize_redirect(request, redirect_uri)
+    return orcid_auth_uri
 
 
 @manage_api.get("/auth")
@@ -216,7 +220,13 @@ async def auth(request: starlette.requests.Request):
     token = await oauth.orcid.authorize_access_token(request)
     request.session["user"] = dict(token)
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    return starlette.responses.RedirectResponse(url=redirect_url)
+    # redirect_url = "http://localhost:3000/#/dois"
+    if "annotation" in request.query_params and request.query_params["annotation"]=="true":
+        # if login is for annotation purpose, add access token as query param 
+        redirect_url += "?access_token=" + token["access_token"]
+        return starlette.responses.RedirectResponse(url=redirect_url) 
+    else:
+        return starlette.responses.RedirectResponse(url=redirect_url)
 
 
 @manage_api.get("/logout")
@@ -246,6 +256,7 @@ def userinfo(request: starlette.requests.Request):
             "id_token": user.get("id_token"),
             "expires_at": user.get("expires_at"),
             "auth_time": auth_time,
+            "access_token" : user.get("access_token")
         }
     else:
         # I think the middleware should prevent this, but just in case…
--- a/isb_web/manage.py
+++ b/isb_web/manage.py
@@ -204,8 +204,8 @@ async def login(request: starlette.requests.Request):
     Initiate OAuth2 login with ORCID
     """
     redirect_uri = request.url_for("auth")
-    # check if login is for annotation purpose, if so add query param 
-    if "annotation" in request.query_params and request.query_params["annotation"]=="true" :
+    # check if login is for annotation purpose, if so add query param
+    if "annotation" in request.query_params and request.query_params["annotation"] == "true":
         redirect_uri += "?annotation=true"
     orcid_auth_uri = await oauth.orcid.authorize_redirect(request, redirect_uri)
     return orcid_auth_uri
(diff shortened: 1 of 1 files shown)


isb_web/manage.py
```diff
@@ -204,7 +204,11 @@ async def login(request: starlette.requests.Request):
     Initiate OAuth2 login with ORCID
     """
     redirect_uri = request.url_for("auth")
-    return await oauth.orcid.authorize_redirect(request, redirect_uri)
+    # check if login is for annotation purpose, if so add query param 
+    if "annotation" in request.query_params and request.query_params["annotation"]=="true" :
```
hyunssong: Using the query parameter, we can distinguish whether we are logging in for throughput annotation purposes.

isb_web/manage.py
```diff
@@ -204,7 +204,11 @@ async def login(request: starlette.requests.Request):
...
+        redirect_uri += "?annotation=true"
```
hyunssong, line 209: If we are accessing the `login` endpoint from throughput, we pass the `annotation=true` query param to the `auth` endpoint as well.

isb_web/manage.py
```diff
@@ -216,7 +220,13 @@ async def auth(request: starlette.requests.Request):
     token = await oauth.orcid.authorize_access_token(request)
     request.session["user"] = dict(token)
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    return starlette.responses.RedirectResponse(url=redirect_url)
+    # redirect_url = "http://localhost:3000/#/dois"
+    if "annotation" in request.query_params and request.query_params["annotation"]=="true":
+        # if login is for annotation purpose, add access token as query param 
+        redirect_url += "?access_token=" + token["access_token"]
```
hyunssong, line 226: At the `auth` endpoint, we check the query params to see if we are accessing for throughput annotation purposes. If so, we append the access token as a query param to the redirect url. This allows us to avoid saving the access token into the cookies.

isb_web/manage.py
```diff
@@ -216,7 +220,13 @@ async def auth(request: starlette.requests.Request):
     token = await oauth.orcid.authorize_access_token(request)
     request.session["user"] = dict(token)
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    return starlette.responses.RedirectResponse(url=redirect_url)
+    # redirect_url = "http://localhost:3000/#/dois"
```
hyunssong, line 223: For testing purposes, I manually specified the redirect_url as localhost.

isb_web/manage.py
```diff
@@ -246,6 +256,7 @@ def userinfo(request: starlette.requests.Request):
             "id_token": user.get("id_token"),
             "expires_at": user.get("expires_at"),
             "auth_time": auth_time,
+            "access_token" : user.get("access_token")
```
hyunssong: Also expose the access token as we need to use this to get a "throughput token". Using this access token, we are able to exchange it to a "throughput token" from the throughput API, which can then be used to create annotations in throughput.

isb_web/manage.py
hyunssong: Going back to the webui(e.g. localhost:3000/#/) page instead of the dois page won't allow us to add the access token as a query param, as it clears all the query params and appends the previous search params. For now, even if we login for annotation purposes redirect to dois page.

isb_web/manage.py
dannymandel: For testing that works, please remove this line before merging.

isb_web/manage.py
dannymandel: Do we need to make a fix to the `webui` side?

dannymandel: It seems like we might want a generic landing page -- as we add more services we'll definitely want to route to more destinations after logging in.

isb_web/manage.py
```diff
@@ -216,7 +220,13 @@ async def auth(request: starlette.requests.Request):
...
     request.session["user"] = dict(token)
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    return starlette.responses.RedirectResponse(url=redirect_url)
+    # redirect_url = "http://localhost:3000/#/dois"
+    if "annotation" in request.query_params and request.query_params["annotation"] == "true":
+        # if login is for annotation purpose, add access token as query param
+        redirect_url += "?access_token=" + token["access_token"]
+        return starlette.responses.RedirectResponse(url=redirect_url)
+    else:
+        return starlette.responses.RedirectResponse(url=redirect_url)
```
dannymandel, line 229: It seems like we can get rid of this `else` statement, right?

dannymandel: Just move the `return` statement outside the `if` block?

isb_web/manage.py
```diff
@@ -204,7 +204,11 @@ async def login(request: starlette.requests.Request):
     Initiate OAuth2 login with ORCID
     """
     redirect_uri = request.url_for("auth")
-    return await oauth.orcid.authorize_redirect(request, redirect_uri)
+    # check if login is for annotation purpose, if so add query param
+    if "annotation" in request.query_params and request.query_params["annotation"] == "true":
```
dannymandel, line 208: I'm pretty sure you can write this more concisely as 

```
if request.query_params.get("annotation") == "true":
```

isb_web/manage.py
```diff
@@ -216,7 +220,13 @@ async def auth(request: starlette.requests.Request):
     token = await oauth.orcid.authorize_access_token(request)
     request.session["user"] = dict(token)
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    return starlette.responses.RedirectResponse(url=redirect_url)
+    # redirect_url = "http://localhost:3000/#/dois"
+    if "annotation" in request.query_params and request.query_params["annotation"] == "true":
```
dannymandel, line 224: I'm pretty sure you can write this more concisely as 

```
if request.query_params.get("annotation") == "true":
```

dannymandel (approved):
Looks good!  A couple questions/comments, but great work!

isb_web/manage.py
hyunssong: I agree. I added some fixes at `throughput-widget` to redirect to the landing page after saving the throughput tokens in the `dois` page.

datadavev (approved):
Looks good. I agree with the couple of points Danny listed

PR #274, opened 2023-05-30
Don't leak database connections!

dannymandel:
On investigation of the monitoring problem I saw this error:

```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/14/3o7r)
```

Basically, we were leaking database connections in the metrics handler.  The fix for this is to properly use the session generator that manages the context and automatically closes it after the work is done.

Fixed one other place where this was occurring as well.

### DIFF
--- a/isb_web/main.py
+++ b/isb_web/main.py
@@ -107,7 +107,9 @@ app.include_router(metrics.router)
 @app.on_event("startup")
 def on_startup():
     dao.connect_sqlmodel(isb_web.config.Settings().database_url)
-    orcid_ids = sqlmodel_database.all_orcid_ids(dao.get_session())
+    session = dao.get_session()
+    orcid_ids = sqlmodel_database.all_orcid_ids(session)
+    session.close()
     # Superusers are allowed to mint identifiers as well, so make sure they're in the list.
     orcid_ids.extend(isb_web.config.Settings().orcid_superusers)
     # The main handler's startup is the guaranteed spot where we know we have a db connection.
--- a/isb_web/metrics.py
+++ b/isb_web/metrics.py
@@ -2,7 +2,8 @@ import logging
 import time
 from typing import Optional
 
-from fastapi import APIRouter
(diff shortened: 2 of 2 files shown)


isb_web/main.py
```diff
@@ -107,7 +107,9 @@
 @app.on_event("startup")
 def on_startup():
     dao.connect_sqlmodel(isb_web.config.Settings().database_url)
-    orcid_ids = sqlmodel_database.all_orcid_ids(dao.get_session())
+    session = dao.get_session()
+    orcid_ids = sqlmodel_database.all_orcid_ids(session)
+    session.close()
```
dannymandel, line 112: This was an existing bug on app startup.  We don't have the dependency injection at that point so just manually open and close the session here.

isb_web/metrics.py
```diff
@@ -45,10 +51,10 @@ def metrics_string(self) -> str:
         return "\n".join(metrics_lines)
 
 
-def _root():
+def _root(session: Session):
     db_start_time = time.time()
     metrics = PrometheusMetrics()
-    metrics.db_counts = things_by_authority_count_dict(dao.get_session())
```
dannymandel, line 51: So this was the cause of the connection leak -- automatically opening up a new db connection on every metrics request and never closing it.  At some point this just exhausts the pool and things go haywire.  The good news is that the alerting was working properly -- I had just introduced this bug when writing the metrics handler.

isb_web/metrics.py
```diff
@@ -14,6 +15,11 @@
 _L = logging.getLogger("metrics")
 
 
+def get_session():
+    with dao.get_session() as session:
```
dannymandel, line 19: This `with` statement ensures that the session is closed when it exits.

dannymandel:
Going to merge this to get it deployed out to see if it fixes the issue as I expect it to.

PR #275, opened 2023-05-31
Disable index creation to see if it helps the postgresql locking

dannymandel:
Previously I had added index creation to ensure they were always there during app startup.  I did see this attempting to execute during the outage period, so I wonder if maybe this could be causing the problem where the locks spike and things slow waaaaay down


dannymandel:
Merging to get this pushed and deployed.

Issue #286, opened 2023-07-28
Make model server invocation use LRU cache


dannymandel:
The iSB model server invocation should use an LRU cache rather than making an http request every time.  I attempted to use it with @functools.cache but you can't use dictionaries with that.  We should be able to hang it on a new method that has already serialized the arguments to bytes and just cache that.

dannymandel:
This is done.

PR #290, opened 2023-08-02
LRU cache for model server

dannymandel:
https://github.com/isamplesorg/isamples_inabox/issues/286


isamples_metadata/taxonomy/metadata_model_client.py
```diff
@@ -29,8 +30,8 @@ def __init__(self, base_url: str, base_headers: dict = {}):
         self.base_url = base_url
         self.base_headers = base_headers
 
-    def _make_json_request(self, url: str, data_params: dict, rsession: requests.Session) -> Any:
-        data_params_bytes: bytes = json.dumps(data_params).encode("utf-8")
+    @functools.lru_cache(maxsize=config.Settings().modelserver_lru_cache_size)
+    def _make_json_request_bytes(self, url: str, data_params_bytes: bytes, rsession: requests.Session) -> Any:
```
dannymandel, line 34: You can't use `@functools.lru_cache` with dictionary arguments since it caches the function args as part of the LRU cache key.  So, cache the raw JSON bytes of the dictionary instead.

PR #292, opened 2023-08-08
Default to https due to SESAR sitemap format change

dannymandel:
On investigation into why we weren't getting SESAR updates, discovered that the SESAR sitemap format had changed and no longer had any scheme specified in their urls.


dannymandel (commented):
This is working on henry…going to merge it.

PR #294, opened 2023-08-08
Login/logout urls were broken on AWS

dannymandel:
Things were working if you tested locally but not in a deployed container.  Fix this.


dannymandel:
Also need to fix the redirect_uri that is generated -- it's generating `http` urls instead of `https`.

dannymandel:
@datadavev FYI, going to push this out to AWS now and verify that hypothesis integration still works.

PR #296, opened 2023-08-11
Oc api update from develop

### LINKED ISSUE #295: Adopt new OpenContext API
The OpenContext records are now vended by a new API:

https://opencontext.org/query/.json?attributes=ALL-STANDARD-LD&cat=oc-gen-cat-sample-col%7C%7Coc-gen-cat-bio-subj-ecofact%7C%7Coc-gen-cat-object&response=metadata%2Curi-meta&sort=updated--desc&type=subjects

We should adopt it and get caught up, as it looks like our last data from OpenContext is roughly ~4/22.

ekansa:
This update is in response to issue: https://github.com/isamplesorg/isamples_inabox/issues/295

This change includes:
(1) Updated the Open Context API URL following updates to Open Context's API
(2) I added a global `OPENCONTEXT_PAGE_SIZE` set to 250 to pull more records from Open Context per request
(3) I made a few modifications to the lists configuring mappings between Open Context "item categories" and iSamples materials and categories
(4) I added a function / method (`_get_oc_str_or_dict_item_label`) for reading either a string value or the 'label' key from a dictionary to get the value for certain attributes. I'm not sure how this should work with the existing Transformer `_transform_key_to_label`  method

### DIFF
--- a/isamples_metadata/OpenContextTransformer.py
+++ b/isamples_metadata/OpenContextTransformer.py
@@ -169,27 +169,25 @@ class OpenContextTransformer(Transformer):
             "late bce/ce", self.source_record, description_pieces
         )
         self._transform_key_to_label("updated", self.source_record, description_pieces)
-        for consists_of_dict in self.source_record.get("Consists of", []):
-            self._transform_key_to_label(
-                "label", consists_of_dict, description_pieces, "Consists of"
+        for consists_of_str in self.source_record.get("Consists of", []):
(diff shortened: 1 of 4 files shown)


dannymandel:
Need to merge this so we can get it deployed out to henry for testing purposes

dannymandel:
We're about 20,000 records in to the reindex, and things appear to be progressing smoothly.  Going to merge this since it seems to be working.

PR #311, opened 2023-09-26
New metadata format

dannymandel:
Update transformers (except SESAR) to the new metadata format


datadavev (approved):
This is really nice work Danny. The PR looks good. While reading about the general problem of record transformation I learned that python 3.10 introduced "pattern matching" syntax which could be leveraged in a few places. See https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

dannymandel:
Going to merge this one so I can make a new PR to use the new OpenContext API.

PR #313, opened 2023-10-10
Implement a couple hacks to work around some of the inconsistencies we have seen with the OpenContext JSON

dannymandel:
@ekansa @datadavev -- FYI, per discussion this morning


ekansa (approved):
Looks good to me! The random foo parameter value seems to always kick in, which will always mean you will bypass our cache. This is nice simple way to deal with our troublesome solr server. I should also make improvements to OC to return a reasonable HTTP error response instead of a null value result.

PR #316, opened 2023-10-17
Wrap the oc request in an exception handler to catch timeouts

dannymandel:
Workaround to handle the timeouts we are seeing


dannymandel:
Manually applied to the OpenContext iSB instance on henry and things are progressing.

dannymandel:
Going to merge this one since it seems to improve the behavior.

PR #319, opened 2023-10-18
Remove heartrate as it is unused

dannymandel:
Causing problems for @SeanSCao when he tries to import the `isb_lib.core` module.


dannymandel:
Merging to hopefully unblock Sean's work.

PR #324, opened 2023-11-03
Don't need these installs, and they are failing on github

dannymandel:
On analyzing some failures in our integration tests, running pip attempting to install these were failing.  We don't actually need either of those to run the integration tests, so remove them.


dannymandel:
Manually ran here: https://github.com/isamplesorg/isamples_inabox/actions/runs/6749284294

It passed, so I'm going to merge to get our automated runs back on track.

PR #325, opened 2023-11-07
Change sitemap generation to happen from database rather than solr

dannymandel:
Since SESAR doesn't yet have a solr index, build the sitemaps from the db instead of solr.


dannymandel:
Currently generating a sitemap on my local machine.  Unfortunately it is quite a bit slower than running this off solr…will probably end up taking ~40 minutes.

PR #328, opened 2023-11-14
Up OC page size per what we've had success with on henry.cyverse.org

dannymandel:
OpenContext is happily returning 1000 rows at a time, so use that…


dannymandel:
Merging to match what we were running on henry

PR #330, opened 2023-11-15
Nonblocking analytics

dannymandel:
Convert the analytics calls to asynchronous middleware


dannymandel:
@datadavev, finished up your good work :)

PR #344, opened 2024-01-09
For Post methods, support querying by including the parameters in the request body if it is encoded as JSON

dannymandel:
As @rdhyee noted, we weren't properly read solr requests out of the body if the method was post (it was looking in the query string).  For post methods, it makes sense to read this out of the request body and expect it to be JSON.  This is what we are using to query solr, and this is what the solr documentation recommends.

### DIFF
--- a/isb_web/main.py
+++ b/isb_web/main.py
@@ -281,11 +281,20 @@ async def _get_solr_select(request: fastapi.Request):
         "q": defparams["q"]
     }
     params = []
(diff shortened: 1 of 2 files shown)


dannymandel:
With this change, @rdhyee's example code now outputs the same with `GET` and `POST`.

dannymandel:
I had to make a couple changes to the sample code to make it work:

image

PR #348, opened 2024-01-26
First stab at improving API documentation

dannymandel:
Polish up the docs in our `openapi.json`.  End result:

image


isb_web/main.py
```diff
@@ -360,15 +381,7 @@ def get_reliquery(request: fastapi.Request, params: ReliqueryParams) -> Reliquer
     return reliquery_response
 
 
-@app.post(f"/{THING_URL_PATH}/select", response_model=typing.Any)
-async def get_solr_query(
-    request: fastapi.Request, query: typing.Any = fastapi.Body(...)
-):
-    # logging.warning(query)
-    return isb_solr_query.solr_query(request.query_params, query=query)
```
dannymandel, line 368: @datadavev This is why the docs weren't showing up in the right place.  We had two solr methods listed, and one of them lacked the necessary docs tags.  Fixed by removing the duplicate.

dannymandel:
@datadavev FYI.  We can polish up further, but this is a big improvement.

PR #354, opened 2024-02-09
Performance improvement -- load vocabulary on app startup and cache it forever

dannymandel:
After seeing performance issues on hyde, we need to make this faster.

### DIFF
--- a/isb_lib/vocabulary/vocab_adapter.py
+++ b/isb_lib/vocabulary/vocab_adapter.py
@@ -14,7 +14,7 @@ import logging
 from term_store import TermRepository
 from term_store.db import Term
 
-VOCAB_CACHE = {}
+VOCAB_CACHE: dict = {}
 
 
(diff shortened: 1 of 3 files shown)


isb_lib/vocabulary/vocab_adapter.py
```diff
@@ -37,11 +38,15 @@ def _read_descendants(term: Term, repository: TermRepository) -> dict:
     return term_dict
 
 
-@functools.lru_cache(maxsize=10)
```
dannymandel, line 40: After turning on SQL logging, it became apparent that this didn't work as intended.  The `repository` parameter is part of the cache key, and it was preventing things from being cached.  Rather than trying to make `TermRepository` be cacheable, it seemed clearer to just make this work out of a dictionary.

dannymandel:
Going to merge this to push this out to hyde to see if it performs better.

PR #355, opened 2024-02-13
Asynchronous iSamples Result exporter

dannymandel:
This is what I get when I hit it locally:

```
(vocabulary) mandeld@Daniels-MacBook-Pro isamples_inabox % curl -v "http://localhost:8000/download/csv?q=foobar"
*   Trying [::1]:8000...
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000
> GET /download/csv?q=foobar HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 201 Created
< date: Tue, 13 Feb 2024 22:40:21 GMT
< server: uvicorn
< content-length: 37
< content-type: application/json
< 
* Connection #0 to host localhost left intact
{"message":"data generation started"}%                                                                                                                                                                                                                                                                                                                                                  (vocabulary) mandeld@Daniels-MacBook-Pro isamples_inabox % cat /tmp/query.txt 
foobar%                           
```


isb_lib/utilities/solr_result_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+        self._solr_result_path = solr_result_path
+        self._format = format
+        self._transformed_result_path = transformed_result_path
+
+    def transform(self):
+        if not os.path.exists(self._solr_result_path):
+            raise ExportTransformException(f"Source solr result file doesn't exist: {self._solr_result_path}")
+        try:
+            with open(self._solr_result_path, "r") as file:
+                solr_result_dict: dict = json.load(file)
```
datadavev: Might run into resource use issue here as this is a potentially large structure being loaded. Might need to consider a streaming parser or something that can load `docs` as individual entries.

isb_web/main.py
```diff
@@ -77,6 +77,7 @@
 manage.dao = dao
 metrics.dao = dao
 vocabulary.dao = dao
+export.dao = dao
```
datadavev, line 80: these global dao vars always seem a bit odd, but it seems to be the way it's done in FastAPI...

datadavev (approved):
This all looks good. There's a couple items that will need to change for production to handle potentially very large results. In general we want to load only chunks of a dataset in memory at a time, so the transformer may need to change a bit to use iterators that load docs one at a time or in chunks rather than all at once. That will require some thinking about the json parsing - perhaps streaming just the docs entries to disk as json lines (i.e. one doc per row and ignoring the solr response prequel) then reading from disk one row at a time which would be one doc at a time.

For now though, this looks just fine for all but the largest result sets.

isb_lib/utilities/solr_result_transformer.py
dannymandel: It looks like `JSONDecoder.raw_decode` might be a way to do this entry by entry.  Not sure if there is a higher level thing we could use that might be a bit easier.

isb_web/main.py
dannymandel: There could be a better way to do it using the dependency injection features of FastAPI, but I haven't really spent time investigating.

isb_lib/utilities/solr_result_transformer.py
dannymandel: This looks promising: https://pypi.org/project/json-stream/

PR #364, opened 2024-04-09
Move to using identifiers in controlled vocabulary

dannymandel:
@datadavev per our discussions this morning, both the migration and the initial generation appear to be working now.  Wrote a test script that picked 1000 random things, ran Transforms on them, and verified the controlled vocabularies were populated.  Flushed out a number of bugs that way.


isamples_metadata/Transformer.py
```diff
@@ -185,29 +186,29 @@ def _rule_based_confidence_list_for_categories_list(category_list: list) -> typi
         return confidences
 
     @abstractmethod
-    def has_context_categories(self) -> typing.List[dict[str, str]]:
+    def has_context_categories(self) -> typing.List[VocabularyTerm]:
```
dannymandel, line 189: All these methods were updated to use real objects instead of dicts.

isamples_metadata/GEOMETransformer.py
```diff
@@ -143,18 +144,18 @@ def sample_description(self) -> str:
 
         return Transformer.DESCRIPTION_SEPARATOR.join(description_pieces)
 
-    def has_context_categories(self) -> list:
+    def has_context_categories(self) -> list[VocabularyTerm]:
         # TODO: resolve https://github.com/isamplesorg/isamples_inabox/issues/312
         # This should probably return the biological kingdom once that is hooked into the vocabulary
-        return [vocabulary_mapper.SAMPLED_FEATURE.term_for_key("sf:marinewaterbody").metadata_dict()]
+        return [vocabulary_mapper.sampled_feature_type().term_for_key("sf:marinewaterbody")]
```
dannymandel, line 150: One annoying fallout of moving things to the database (these used to reference the text file dumps @smrgeoinfo provided me), is that we can't reference static variables.  Now, a pedantic programmer might point out that static variables are a poor design choice, and they wouldn't be wrong, but it *was* simpler before.  :)

dannymandel: So, the replacement is a function that fetches them from the database (or if they've been fetched already, out of a cache).

isamples_metadata/SmithsonianTransformer.py
```diff
@@ -133,18 +139,18 @@ def has_context_categories(self) -> typing.List[dict[str, str]]:
                 self.source_record.get("higherClassification", ""),
             ]
         )
-        return [vocabulary_mapper.SAMPLED_FEATURE.term_for_label(category).metadata_dict() for category in categories]
+        return [vocabulary_mapper.sampled_feature_type().term_for_label(category)]
```
dannymandel, line 142: This line was the source of the odd "a", "b", "c", "d", "e" we were seeing in the index for the Smithsonian records' controlled vocabularies.  When we converted things over to use the modelserver endpoint, the client code thought it was receiving a `list` but the modelserver is actually returning a `str`.  I've updated the API to reflect this.

isamples_metadata/SmithsonianTransformer.py
```diff
@@ -185,8 +191,8 @@ def produced_by_feature_of_interest(self) -> str:
         return Transformer.NOT_PROVIDED
 
     def _add_to_responsibilities(self, source_label: str, dict_label: str, responsibilities: list[dict[str, str]]):
-        value = self.source_record[source_label]
-        if len(value) > 0:
+        value = self.source_record.get(source_label)
+        if value is not None and len(value) > 0:
```
dannymandel, line 195: These were causing crashes for me running locally -- not sure why we hadn't seen them before, but I went ahead and fixed it since it was straightforward and safe.

isamples_metadata/SmithsonianTransformer.py
```diff
@@ -196,8 +202,8 @@ def produced_by_responsibilities(self) -> list[dict[str, str]]:
         self._add_to_responsibilities("scientificNameAuthorship", "scientific name authorship", responsibilities)
 
         # unfortunately it looks like this field uses a ; separator for multiple people
-        identified_by = self.source_record["identifiedBy"]
-        if len(identified_by) > 0:
+        identified_by = self.source_record.get("identifiedBy")
+        if identified_by is not None and len(identified_by) > 0:
```
dannymandel, line 206: These were causing crashes for me running locally -- not sure why we hadn't seen them before, but I went ahead and fixed it since it was straightforward and safe.

isamples_metadata/vocabularies/vocabulary_mapper.py
```diff
@@ -29,32 +29,92 @@ def metadata_dict(self) -> dict[str, str]:
...
-                uri = os.path.join(uri_prefix, key)
-                term = VocabularyTerm(key, label, uri)
-                self.vocabulary_terms_by_key[key] = term
-                self.vocabulary_terms_by_label[label] = term
+    def __init__(self, uijson_dict: dict[str, Any], key_prefix: str):
+        self.vocabulary_terms_by_key: dict[str, VocabularyTerm] = {}
+        self.vocabulary_terms_by_label: dict[str, VocabularyTerm] = {}
+        self._key_prefix = key_prefix
+        self._is_first = True
+        self._process_uijson_dict(uijson_dict)
```
dannymandel, line 38: Left the previous class in place, but swapped out the guts underneath it to operate on the JSON that we generate from fetching everything from the database.

isb_lib/vocabulary/vocab_adapter.py
```diff
@@ -11,13 +11,17 @@
...
 from term_store import TermRepository
 from term_store.db import Term
 
 VOCAB_CACHE: dict = {}
 
 
-def _read_descendants(term: Term, repository: TermRepository) -> dict:
+def _read_descendants(term: Term, repository: TermRepository, all_uris: set) -> Optional[dict]:
+    if term.uri in all_uris:
+        return None
```
dannymandel, line 24: This works around the bug we saw where we were getting `Row` instead of `Term` which caused things to be visited multiple times.

isb_web/sqlmodel_database.py
```diff
@@ -206,6 +206,11 @@ def get_thing_with_id(session: Session, identifier: str) -> Optional[Thing]:
     return result
 
 
+def random_things_with_authority(session: Session, authority: str, count: int) -> list[Thing]:
+    statement = select(Thing).where(Thing.authority_id == authority).limit(count).order_by(func.random())
+    return session.exec(statement).all()
```
dannymandel, line 211: Useful for testing!

PR #366, opened 2024-04-16
A couple small updates based on local testing…

dannymandel:
Ran into a couple issues trying to follow the old documentation and mint new identifiers in DataCite.

### DIFF
--- a/docs/authentication_and_identifiers.md
+++ b/docs/authentication_and_identifiers.md
@@ -111,7 +111,7 @@ Once all of the above has been completed, you're ready to mint new identifiers w
 
 (1) Using the management UI, hit the `/dois` URL in the ISB instance e.g. on [Mars](https://mars.cyverse.org/isamples_central/ui/#/dois) and enter the required fields.  You can see the generated JSON in the right-side panel, and that generated JSON is sent to datacite.  Once the datacite response comes back, you'll be able to see the newly minted identifier.
 
-(2) Similarly, you can also use a CLI to directly hit the iSB `/manage/mint_draft_identifiers` endpoint and mint identifiers using your scripting language of choice.  The only prerequisite here is to make sure you've obtained a valid JWT from orcid before calling the endpoint.  You'll need to include the JWT in an `authorization: Bearer` header when you call the endpoint.  A working example of a CLI is available on the [iSamples GitHub](https://github.com/isamplesorg/isamples_inabox/blob/develop/scripts/examples/mint_identifiers_cli.py)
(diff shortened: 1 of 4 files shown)


scripts/examples/mint_identifiers_cli.py
```diff
@@ -17,13 +17,14 @@
...
     type=str,
-    default="https://mars.cyverse.org/isamples_central/manage/mint_draft_datacite_identifiers",
+    default="https://central.isample.xyz/isamples_central/manage/mint_draft_datacite_identifiers",
     help="The url to the mint identifiers endpoint",
 )
 identifiers_option = click.option(
     "-n",
     "--num_identifiers",
     type=str,
+    default="1",
```
dannymandel, line 27: Things totally break if this isn't specified, and the error message isn't helpful at all, so default it to 1.

docs/authentication_and_identifiers.md
```diff
@@ -111,7 +111,7 @@ Once all of the above has been completed, you're ready to mint new identifiers w
...
 
-(2) Similarly, you can also use a CLI to directly hit the iSB `/manage/mint_draft_identifiers` endpoint and mint identifiers using your scripting language of choice.  The only prerequisite here is to make sure you've obtained a valid JWT from orcid before calling the endpoint.  You'll need to include the JWT in an `authorization: Bearer` header when you call the endpoint.  A working example of a CLI is available on the [iSamples GitHub](https://github.com/isamplesorg/isamples_inabox/blob/develop/scripts/examples/mint_identifiers_cli.py)
+(2) Similarly, you can also use a CLI to directly hit the iSB `/manage/mint_draft_datacite_identifiers` endpoint and mint identifiers using your scripting language of choice.  The only prerequisite here is to make sure you've obtained a valid JWT from orcid before calling the endpoint.  You'll need to include the JWT in an `authorization: Bearer` header when you call the endpoint.  A working example of a CLI is available on the [iSamples GitHub](https://github.com/isamplesorg/isamples_inabox/blob/develop/scripts/examples/mint_identifiers_cli.py)
```
dannymandel, line 114: The API endpoint got renamed when we introduced noidy minting, but we didn't update the docs.

isb_web/config.py
```diff
@@ -39,7 +39,7 @@ class Settings(BaseSettings):
     # This shouldn't be checked in.  Set by doing export DATACITE_PASSWORD=foobar
     datacite_password: str = ""
 
-    orcid_hostname: str = "pub.sandbox.orcid.org"
+    orcid_hostname: str = "orcid.org"
```
dannymandel, line 42: This is only used for local development, but there's no reason to be using the sandbox anymore.

PR #376, opened 2024-04-24
Handle errors when sending solr requests

dannymandel:
Fix the bug that @smrgeoinfo found when attempting to use the export service.  It's very easy to send a malformed query to solr that results in a 400 error.  We now respond appropriately, and the export client CLI also handles these responses appropriately:

```
WARNING:root:Export job failed with error.  Check that your solr query is valid and try again.  Response: {'status': 'error', 'tcompleted': '2024-04-24 14:19:31.131047', 'reason': 'HTTP Error, code: 400 reason: Bad Request'}
```

### DIFF
--- a/isamples_export_client/export_client.py
+++ b/isamples_export_client/export_client.py
@@ -4,7 +4,7 @@ import logging
 import time
 from enum import Enum
 import os
-from typing import Optional
+from typing import Optional, Any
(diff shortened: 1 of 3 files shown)


dannymandel:
Going to merge this one so we can push it out.

PR #382, opened 2024-05-20
Export service json validator fixes

dannymandel:
Tweaks to get the exported JSON lines passing the json schema.  Also include the validator check as part of the test so we ensure we are always passing.


tests/test_data/json_schema/iSamplesSchemaCore1.0.json
```diff
@@ -235,7 +235,15 @@
                 },
                 "result_time": {
                     "description": "Date on which the sample was collected.",
-                    "format": "date",
+                    "anyOf": [
+                        {
+                            "format": "date"
+                        },
+                        {
+                            "format": "date-time"
```
dannymandel, line 243: I'm not certain this is correct.  It passes, though.

dannymandel:
The export service is now passing the JSON schema validator!

tests/test_data/json_schema/iSamplesSchemaCore1.0.json
dannymandel: @datadavev if you have thoughts here this is what I'm unsure about

dannymandel: Per discussion with @datadavev in this morning's standup, this is correct: https://json-schema.org/understanding-json-schema/reference/string#built-in-formats

dannymandel:
Going to merge this one.

Issue #383, opened 2024-05-22
Encode content with json.dumps() instead os str()


datadavev:
https://github.com/isamplesorg/isamples_inabox/blob/1dc31e4feb27e06b7ef5d62996f9a52c57da54e2/isb_web/export.py#L150

The export client failed to properly handle a solr request generated by the ui:
```
{'q': 'hasMaterialCategory:(\"Fluid material\" OR \"Gaseous material\" OR \"Liquid water\" OR \"Non-aqueous liquid material\") AND -relation_target:*', 'wt': 'json', 'fl': 'id,authorizedBy,compliesWith,producedBy_samplingSite_location_longitude,producedBy_samplingSite_location_latitude,relatedResource_isb_core_id,curation_responsibility,curation_location,curation_accessContraints,curation_description,curation_label,samplingPurpose,registrant,producedBy_samplingSite_placeName,producedBy_samplingSite_location_elevationInMeters,producedBy_samplingSite_label,producedBy_samplingSite_description,producedBy_resultTime,producedBy_responsibility,producedBy_hasFeatureOfInterest,producedBy_description,producedBy_label,producedBy_isb_core_id,informalClassification,keywords,hasSpecimenCategory,hasMaterialCategory,hasContextCategory,description,label,source', 'sort': 'id asc'}
```
reporting `json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)`

The culprit is the opening parenthesis at `hasMaterialCategory:(\"Fluid material\" ...` 

The status response generator should use: 
```
"query": json.dumps(export_job.solr_query_params)
```
instead of:
```
"query": str(export_job.solr_query_params)
```
to properly encode the query params to serialize as a json string within the status json response.

The export client can then use:
```
query_string = json.loads(status_json.get("query"))
```
instead of:
```
query_string = status_json.get("query").replace("'", "\"")
```

dannymandel:
This will be fixed once we do our next push to central

PR #385, opened 2024-05-29
Fix for https://github.com/isamplesorg/isamples_inabox/issues/383

dannymandel:
Properly encode JSON query params


dannymandel:
Tested out locally using the iSamples Export Client, results look identical.

Issue #372, opened 2024-04-18
Global Grid cell identifiers are  included in core JSON API response


smrgeoinfo:
e.g. 
```
   "producedBy_samplingSite_location_h3_0": "8045fffffffffff",
    "producedBy_samplingSite_location_h3_1": "8144bffffffffff",
    "producedBy_samplingSite_location_h3_2": "8244affffffffff",
    "producedBy_samplingSite_location_h3_3": "8344aefffffffff",
    "producedBy_samplingSite_location_h3_4": "8444ae3ffffffff",.....
```

These aren't in the schema and I don't think many people will have a use for them. I'd recommend not inserting them in the JSON.

dannymandel:
This is an easy fix.

dannymandel:
This will be fixed once we do our next push to central

PR #386, opened 2024-05-29
Don't include h3 values in JSON 

dannymandel:
This issue says no h3 in JSON: https://github.com/isamplesorg/isamples_inabox/issues/372


datadavev (approved):
I think this approach makes sense

PR #394, opened 2024-06-18
Add a link to the original record

dannymandel:
Since we are now linking iSamples records in the list to the iSamples page, make sure that page has a link to the original record.

image


dannymandel:
Going to merge and deploy this along with a few other fixes

PR #395, opened 2024-06-26
Sitemap generation through the export service.

dannymandel:
Sitemap generation through the export service.


isb_web/auth.py
```diff
@@ -82,10 +83,30 @@ async def __call__(
             scope["oauth2-claims"] = claims
             scope["oauth2-provider"] = provider
             scope["oauth2-jwt"] = token
-        except starlette_oauth2_api.InvalidToken as e:
-            return await self._prepare_error_response(
-                e.errors, 401, scope, receive, send
-            )
+        except starlette_oauth2_api.InvalidToken:
```
dannymandel, line 86: The one thing I am not crazy about with this is that we don't know beforehand if it's a JWT or an access token.  So the flow is "check JWT, then if it fails check access token".  If you've got a better way of trying this, @datadavev I'd love to hear it.

isb_web/manage.py
```diff
@@ -112,9 +114,11 @@ async def auth(request: starlette.requests.Request):
     token_dict = dict(token)
     request.session["user"] = token_dict
     redirect_url = url_utilities.joined_url(str(request.url), config.Settings().auth_response_redirect_fragment)
-    # redirect_url = "http://localhost:3000/#/dois"
+
     if "raw_jwt" in request.query_params and request.query_params["raw_jwt"] == "true":
         return PlainTextResponse(content=token_dict["id_token"])
+    elif "raw_token" in request.query_params and request.query_params["raw_token"] == "true":
+        return PlainTextResponse(content=token_dict["access_token"])
```
dannymandel, line 121: This will echo the `access_token` instead of the JWT to the response.  I left the JWT path in the code since it seemed like it could potentially be useful, too.

dannymandel:
On reviewing some of the other export bits, I think we need more than this, actually.  In the export service, we record the orcid of the user that initiated the download, and that code reads that out of the JWT.  We'll need to change that code as well, to do a lookup against the user info service in case a token is used.

dannymandel:
Going to merge this one and deploy out to Henry for testing

PR #399, opened 2024-07-17
One last sitemap fix to get the url to the JSON lines right

dannymandel:
Fix the urls to the right paths


dannymandel:
@datadavev FYI

PR #403, opened 2024-08-07
Move superusers to a secret that is exported via the environment

dannymandel:
Simple change on this side, will be some more work on the isamples_docker side


dannymandel:
Going to merge this one so I can test it out on henry

PR #404, opened 2024-08-07
Update the superuser config docs per what actually is working on henry

dannymandel:
Surprisingly this was working differently there -- I think I need to update my local venv


dannymandel (commented):
@datadavev FYI

PR #407, opened 2024-08-28
Have SESAR write out a sitemap index and entries with json lines files

dannymandel:
Don't worry about solr running in the SESAR setup.

### DIFF
--- a/isb_lib/sitemaps/sesar_thing_sitemap.py
+++ b/isb_lib/sitemaps/sesar_thing_sitemap.py
@@ -6,8 +6,7 @@ from typing import Optional
 from sqlmodel import Session
 
 from isb_lib.core import datetimeToSolrStr
-from isb_lib.sitemaps import SitemapIndexEntry, UrlSetEntry, ThingUrlSetEntry, ThingSitemapIndexEntry, \
-    SESARThingUrlSetEntry
+from isb_lib.sitemaps import SitemapIndexEntry, UrlSetEntry, ThingSitemapIndexEntry, SESARThingUrlSetEntry
 from isb_web.sqlmodel_database import things_for_sitemap
 
 MAX_URLS_IN_SITEMAP = 50000
@@ -49,7 +48,7 @@ class SESARThingUrlSetIterator:
 
     def sitemap_index_entry(self) -> SitemapIndexEntry:
         return ThingSitemapIndexEntry(
-            f"sitemap-{self.sitemap_index}.xml", self.last_tstamp_str or ""
+            f"sitemap-{self.sitemap_index}.jsonl", self.last_tstamp_str or ""
         )
 
 
--- a/isb_lib/sitemaps/sesar_thing_sitemap.py
+++ b/isb_lib/sitemaps/sesar_thing_sitemap.py
@@ -1,0 +1,96 @@
+
+import datetime
+import typing
+from typing import Optional
+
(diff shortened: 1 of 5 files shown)


dannymandel:
I ran with these parameters locally:

```
-p /tmp/sitemaps/ -h https://central.isample.xyz/ -o 5000000
```

And this is the output:

```
mandeld@Daniels-MBP isamples_inabox % ls -la /tmp/sitemaps 
total 306808
drwxr-xr-x  5 mandeld  wheel        160 Aug 28 15:25 .
drwxrwxrwt  9 root     wheel        288 Aug 28 12:20 ..
-rw-r--r--@ 1 mandeld  wheel  113494749 Aug 28 15:25 sitemap-0.jsonl
-rw-r--r--@ 1 mandeld  wheel   38399667 Aug 28 15:25 sitemap-1.jsonl
-rw-r--r--@ 1 mandeld  wheel        587 Aug 28 15:25 sitemap-index.xml

mandeld@Daniels-MBP isamples_inabox % cat /tmp/sitemaps/sitemap-index.xml 
<?xml version="1.0" encoding="utf-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 https://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd">
  <sitemap>
    <loc>https://central.isample.xyz/sitemaps/sitemap-0.jsonl</loc>
    <lastmod>2024-02-23T08:51:58.729367Z</lastmod>
  </sitemap>
  <sitemap>
    <loc>https://central.isample.xyz/sitemaps/sitemap-1.jsonl</loc>
    <lastmod>2024-04-09T23:55:09.354934Z</lastmod>
  </sitemap>
</sitemapindex>%   
```

I won't include the output from the .jsonl files here, because they're as expected.

dannymandel:
So @SeanSCao you'd just want to use whatever paths for `-h` and `-p` that would correspond to the URL where you'll host the sitemap and the path in the Docker image where you'll want the files written out to.

You can ignore the `-o` -- I was just doing that so it would start at index 5000000 and only write a couple files so I could check that the offsets were working and have it complete quickly.

isb_lib/sitemaps/sesar_thing_sitemap.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import Optional
+
+from sqlmodel import Session
+
+from isb_lib.core import datetimeToSolrStr
+from isb_lib.sitemaps import SitemapIndexEntry, UrlSetEntry, ThingSitemapIndexEntry, SESARThingUrlSetEntry
+from isb_web.sqlmodel_database import things_for_sitemap
+
+MAX_URLS_IN_SITEMAP = 50000
+
```
dannymandel, line 13: This file was restored from a previously deleted version.  The SESAR prefix was included to avoid confusion.  All other iSB instances just use the export handler now which has a very different mechanism for writing the .jsonl files.

dannymandel:
@SeanSCao @datadavev after thinking about our conversation at standup and reviewing the code in this PR, I think that iSB is the right place for this code.  This script should run inside the box itself, so iSB is the right place for it.

Issue #411, opened 2024-09-04
GEOME solr indexer wasn't updating, due to uninitialized vocabulary caches


dannymandel:
When checking on the GEOME solr indexer today, I noticed it was broken due to the vocabulary caches being uninitialized.  I got it running on Henry with this diff:

```
diff --git a/scripts/geome_things.py b/scripts/geome_things.py
index b40be67..34aab30 100644
--- a/scripts/geome_things.py
+++ b/scripts/geome_things.py
@@ -4,6 +4,7 @@ import time
 from concurrent.futures import Future
 from typing import Optional
 
+import term_store
 import requests
 import sqlalchemy
 import sqlalchemy.orm
@@ -22,6 +23,10 @@ from isb_lib.models.thing import Thing
 from isb_web import sqlmodel_database
 from isb_web.sqlmodel_database import SQLModelDAO, get_thing_with_id, save_thing, all_thing_primary_keys, \
     DatabaseBulkUpdater, get_things_with_ids
+from isb_lib.vocabulary import vocab_adapter
+from isamples_metadata.vocabularies import vocabulary_mapper
+
+from isb_web.vocabulary import SAMPLEDFEATURE_URI, MATERIAL_URI, MATERIALSAMPLEOBJECTTYPE_URI
 
 CONCURRENT_DOWNLOADS = 10
 BACKLOG_SIZE = 40
@@ -350,6 +355,15 @@ def populateIsbCoreSolr(ctx, ignore_last_modified: bool):
         solr_url=solr_url,
         min_time_created=max_solr_updated_date
     )
+    dao = SQLModelDAO(db_url)
+    session = dao.get_session()
+    repository = term_store.get_repository(session)
+    vocab_adapter.uijson_vocabulary_dict(SAMPLEDFEATURE_URI, repository)
+    vocab_adapter.uijson_vocabulary_dict(MATERIAL_URI, repository)
+    vocab_adapter.uijson_vocabulary_dict(MATERIALSAMPLEOBJECTTYPE_URI, repository)
+    vocabulary_mapper.sampled_feature_type()
+    vocabulary_mapper.material_type()
+    vocabulary_mapper.specimen_type()
     allkeys = solr_importer.run_solr_import(isb_lib.geome_adapter.reparseAsCoreRecord)
     logger.info(f"Total keys= {len(allkeys)}")
```

Definitely need to check the other updaters to see if they're broken in the same way.

dannymandel:
[...]

PR #412, opened 2024-09-06
Fix vocabulary term initialization

dannymandel:
https://github.com/isamplesorg/isamples_inabox/issues/411

Since these are in the database (and cached) now, we need to initialize them when we start up.


dannymandel:
Tested out locally, fixes this issue.  While testing locally, discovered another issue: https://github.com/isamplesorg/isamples_inabox/issues/413

PR #423, opened 2024-09-13
Fix sitemap ingest (and writing)

dannymandel:
A few bugs I discovered when testing this out locally using the deployed sitemaps on Henry


dannymandel:
@datadavev FYI

isb_lib/sitemaps/sitemap_fetcher.py
```diff
@@ -111,7 +111,7 @@ def _fetch_file(self):
             lastmod_date = isb_lib.core.parsed_datetime_from_isamples_format(lastmod)
             if (
                 self._last_modified is None
-                or lastmod_date.timestamp() >= self._last_modified.timestamp()
+                or lastmod_date.timestamp() > self._last_modified.timestamp()
```
dannymandel, line 114: This was causing us to always refetch the last file…

isb_web/export.py
```diff
@@ -174,7 +174,7 @@ async def create_sitemap(request: fastapi.Request, session: Session = Depends(ge
     orcid_id = auth.orcid_id_from_session_or_scope(request)
     if orcid_id not in isb_web.config.Settings().orcid_superusers:
         raise HTTPException(401, "orcid id not authorized to create sitemap")
-    solr_api_defparams = _default_export_params("indexUpdatedTime asc")
+    solr_api_defparams = _default_export_params("sourceUpdatedTime asc")
```
dannymandel, line 177: This change is non-trivial.  The implication is that if we were to, say, reindex a bunch of stuff on the iSB instance, we wouldn't get it by default, because the `sourceUpdatedTime` wouldn't be changed.  In a world where we don't own the boxes, this would be up to the provider to determine, so I think it's the right thing to do in the meantime, just FYI.

scripts/consume_sitemaps.py
```diff
@@ -186,7 +186,6 @@ def fetch_sitemap_files(authority, last_updated_date, thing_ids: typing.Dict[str
                             thing_dict["primary_key"] = thing_ids[thing_identifier]
                             current_existing_things_batch.append(thing_dict)
                         else:
-                            thing_dict["tcreated"] = now
```
dannymandel, line 189: This was nuking the previous code that we added that actually tracked the last mod time.  Whoopsie.

scripts/consume_sitemaps.py
```diff
@@ -217,7 +216,7 @@ def _json_line_to_thing_dict(json_dict: dict, json_lines_url: str, now: datetime
     thing_dict["resolved_status"] = 200
     thing_dict["tresolved"] = time_fetched
     thing_dict["resolved_media_type"] = MEDIA_JSONL
-    thing_dict["tcreated"] = json_dict["last_modified_time"]
+    thing_dict["tcreated"] = json_dict[METADATA_LAST_UPDATED_TIME]
```
dannymandel, line 219: And this is why we use constants to refer to strings…=\

datadavev (approved):

PR #424, opened 2024-09-17
Last modified time

dannymandel:
This was the constant that we agreed to use, I mistakenly named it incorrectly because of the API we had in place previously.  There is a bit of tension here now, but in terms of the public facing API (the JSON schema), I think the name is correct and it's more important that is right than any previous internal names we had.


dannymandel:
Going to merge this to Henry to reharvest things with the right column name.  @datadavev FYI

PR #425, opened 2024-09-17
Use the right mapping for source_updated_time, so we don't reindex everything all the time

dannymandel:
Fix for the bug we discussed in standup this morning.  @datadavev FYI


dannymandel:
Going to merge this one (though it only affects the indexer so we don't need to deploy anyway since the only place this is running is on my machine right now)

PR #426, opened 2024-09-17
This fixes the few records reindexing we kept seeing 

dannymandel:
It's unclear to me that this is the right solution, but it does seem to do the trick.  I'm not sure if postgresql is omitting the tz, or if including the "Z" timezone affects the db filtering, but this works around the issue.

I'm hesitant to fully recommend taking the change because:

(1) The effects of reindexing are incredibly minor (extra CPU usage)
(2) It doesn't feel right -- the right solution feels like including the timezone in the database in the same way that solr is doing it (I think solr is including the Z)

@datadavev FYI, I leave this up to you to decide if you want it or not

### DIFF
--- a/isb_lib/core.py
+++ b/isb_lib/core.py
@@ -500,7 +500,8 @@ def solr_max_source_updated_time(
         dict = res.json()
         docs = dict["response"]["docs"]
         if docs is not None and len(docs) > 0:
-            return dateparser.parse(docs[0]["sourceUpdatedTime"])
+            # Strip out the time zone since that is confusing matters
+            return dateparser.parse(docs[0]["sourceUpdatedTime"]).replace(tzinfo=None)
     except Exception:
         getLogger().error("Didn't get expected JSON back from %s when fetching max source updated time for %s", _url, authority_id)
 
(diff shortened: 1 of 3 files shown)


isb_web/sqlmodel_database.py
```diff
@@ -278,7 +278,7 @@ def paged_things_with_ids(
     thing_select = _base_thing_select(authority, status, limit, offset, min_id)
 
     if min_time_created is not None:
-        thing_select = thing_select.filter(Thing.tcreated >= min_time_created)
+        thing_select = thing_select.filter(Thing.tcreated > min_time_created)
```
dannymandel, line 281: FWIW, this line seems less controversial.  I think I included it because I wasn't sure how accurate dates would be in the database, but it seems to do what we want even if it is limited to `>`.

isb_lib/core.py
```diff
@@ -500,7 +500,8 @@ def solr_max_source_updated_time(
         dict = res.json()
         docs = dict["response"]["docs"]
         if docs is not None and len(docs) > 0:
-            return dateparser.parse(docs[0]["sourceUpdatedTime"])
+            # Strip out the time zone since that is confusing matters
+            return dateparser.parse(docs[0]["sourceUpdatedTime"]).replace(tzinfo=None)
```
dannymandel: This is the line that feels dirty.

dannymandel: Per discussion this morning in standup, the bug was actually in the postgresql instance on my local machine.  It was configured to use the Arizona timezone.  When I changed my local server to run in UTC, everything worked fine.

